PR #37, opened 2025-07-11
don't assume min replicas doesn't update in e2e

paulgmiller:


test/e2e/e2e_test.go
```diff
@@ -361,7 +361,8 @@ var _ = Describe("controller", Ordered, func() {
 			}
 			EventuallyWithOffset(1, verifyMinAvailableUpdated, time.Minute, time.Second).Should(Succeed())
 
-			By("Verify Eviction Autoscaler MinAvailable is not updated")
+			//Autoscalers min available status is not updaed by pdb watcher but can self update if controller restarts/resyncs
+			/*By("Verify Eviction Autoscaler MinAvailable is not updated")
```
skuchipudi295, line 365: just remove this block instead?

PR #40, opened 2025-07-16
helm charts

skuchipudi295:
1. added helm chart, values.yaml 
2.  made all yawls refer values.yaml
3. need to point image to mar later
4. added config map for user configuration 

Configuration Sections

1. Metrics Configuration

metrics:  enabled: true  # Enable/disable Prometheus metrics collection
Purpose: Controls whether the controller exports metrics to Prometheus
Independent: Metrics collection works independently of namespace filtering

2. Namespace Filtering

namespaces:  watchMode: "all"        # Modes: "all", "only", "none"  excludeList: []         # Namespaces to exclude (when watchMode="all")  includeList: []         # Namespaces to include (when watchMode="only")
Watch Modes:

"all": Watch all namespaces, optionally exclude specific ones
"only": Watch only namespaces specified in includeList
"none": Disable namespace watching (controller becomes passive)

3. PDB Configuration

pdb:  create: true  # Enable automatic PodDisruptionBudget creation


helm/eviction-autoscaler/templates/configmap.yaml
```diff
@@ -0,0 +1,28 @@
...
+      enabled: {{ .Values.controllerConfig.metrics.enabled }}
+
+    namespaces:
+      watchMode: {{ .Values.controllerConfig.namespaces.watchMode | quote }}
+{{- if and .Values.controllerConfig.namespaces.excludeList (gt (len .Values.controllerConfig.namespaces.excludeList) 0) }}
+      excludeList:
+{{ toYaml .Values.controllerConfig.namespaces.excludeList | indent 8 }}
+{{- end }}
+{{- if and .Values.controllerConfig.namespaces.includeList (gt (len .Values.controllerConfig.namespaces.includeList) 0) }}
+      includeList:
```
paulgmiller, line 23: I _think_ we may want to do these as annotations on the namepaces themselves.  But why do I think that

Maybe so that webhook can sutomatically turn this on? But why not use excludes.

helm/eviction-autoscaler/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,52 @@
+{{- if and .Values.metrics.enabled .Values.metrics.serviceMonitor.enabled }}
```
paulgmiller, line 1: not actually used by us today. pull it out?

helm/eviction-autoscaler/templates/configmap.yaml
```diff
@@ -0,0 +1,28 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: {{ include "eviction-autoscaler.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+data:
+  config.yaml: |-
+    metrics:
+      enabled: {{ .Values.controllerConfig.metrics.enabled }}
```
Javier090, line 14: prob nitpic but cant this check lead to mismatch? like if this flag is set 0 and Values.metrics.enabled is enabled the controller now has no endpoint but the service endpoint still scraps a now empty endpoint?

Javier090 (approved):
looks fine It dosent looks liek anything metrics should break just left one comment/ more quesiton

helm/eviction-autoscaler/templates/configmap.yaml
Javier090: sorry was a bit wordy, I mean tlike in the two configuratioms if the controller is started with metric-port=0 or  disabled but values.metrics.enabled is set to true, then the service would expect metrics and try to scrape them but the controller isnt providing them, so couldnt those settings be in sync?

skuchipudi295: I think we should not allow metric-port to be 0. metrics collection can only controlled by enabling/disbaling ,values.metrics.enabled

PR #42, opened 2025-07-21
Helme2e

Javier090:
Modified the e2e test to deploy test husing helm. Made changes to helm also for resolving namespaces, and removing unused configuration options.

### Namespace Handling Improvements:
* Introduced a helper function `eviction-autoscaler.ns` to dynamically resolve the namespace for resources, defaulting to `.Release.Namespace` if `.Values.namespace` is not explicitly set (`helm/eviction-autoscaler/templates/_helpers.tpl`).
* Updated all namespace references in resource templates (e.g., `deployment.yaml`, `serviceaccount.yaml`, `metrics-service.yaml`, etc.) to use the new helper function instead of `.Values.namespace`. [[1]](diffhunk://#diff-698aab4cef18204d0f22b56bcbc3bb0d67a950859fb20bb57b8408ce3f1a3701L114-R114) [[2]](diffhunk://#diff-5f139f5160ac21191404843a50aaf7c8f086f3b37f25f1aee8c8c49c5248da54L5-R5) [[3]](diffhunk://#diff-3e0305ce4dd84ef4793d1ac95ce4728c04f97d687eca8947d980d62487a256efL5-R5) [[4]](diffhunk://#diff-3e0305ce4dd84ef4793d1ac95ce4728c04f97d687eca8947d980d62487a256efL48-R48) [[5]](diffhunk://#diff-32c74595af5e55cade734cf0c7acd57c783b9efac38a45a2733a6d5e77ad3df8L5-R5) [[6]](diffhunk://#diff-9b683605746526e9bad77d0fef9ad0adbd5d8f926636b020199b179dfba24f14L5-R5) [[7]](diffhunk://#diff-729bf66e8b163b13195c223ff76175de78b7e4b794b4fb650a1117df9014d64eL5-R5) [[8]](diffhunk://#diff-7a4931d7568a7ada641925430df98c72acbd7673713fde049c0236c1d6c6c675R1-R6)
* Modified the `namespace.yaml` template to exclude namespaces starting with `kube-` from being created automatically.

### Configuration Simplifications:
* Removed the `namespace` field from `values.yaml` and replaced it with an empty default value, as namespaces are now resolved dynamically.
* Removed unused configuration options, including the checksum annotation for the `configmap.yaml` and the volume/volumeMount for the configuration file in the `deployment.yaml`. [[1]](diffhunk://#diff-5f139f5160ac21191404843a50aaf7c8f086f3b37f25f1aee8c8c49c5248da54L22) [[2]](diffhunk://#diff-5f139f5160ac21191404843a50aaf7c8f086f3b37f25f1aee8c8c49c5248da54L68-L75)
* Updated the `ServiceMonitor` template to allow customization of scrape intervals and timeouts via `values.yaml`.

### E2E Testing Updates:
* Modified e2e with Helm for deploying the controller-manager in E2E tests

### Removed kustomize files
* Modified make file and deployment install to use helm instead of kustomize 

### AI Usage
* Used AI to map out e2e tests using helm chart and write pseudocode
* AI used to automate small changes across files quickly such as changing one line
* Used Copilto to explain kind usage with helm and give me documentation
*


hack/install.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
-#So ideally we move towards kustomize and cert manager
-#that will make image and namespace simpler first then cert manger should make the webhook ca bundle easier.
+# Installation script for eviction-autoscaler
```
paulgmiller, line 3: would be fin just deleting this install.sh I haven't used it in a while.

helm/eviction-autoscaler/templates/_helpers.tpl
```diff
@@ -42,6 +42,15 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end }}
 
+{{/*
+Resolve the namespace where namespaced resources should be created.
```
paulgmiller, line 46: not sure I understand this. why woudln't we just use .Release.Namespace

helm/eviction-autoscaler/templates/servicemonitor.yaml
```diff
@@ -1,8 +1,9 @@
+{{- if .Values.serviceMonitor.enabled }}
```
paulgmiller, line 1: Alternatively get rid of this for now.  Maybe some people are using this but it introduces a dependency on having hte monitoring.coreos.com/v1 crd there so not sure we'll ever turn it on

paulgmiller: @skuchipudi295 fyi

helm/eviction-autoscaler/templates/namespace.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if not (eq .Release.Namespace "default") }}
+{{- if and (not (eq .Release.Namespace "default")) (not (hasPrefix "kube-" .Release.Namespace)) }}
```
paulgmiller, line 1: say what you're doing witha  comment>

helm/eviction-autoscaler/values.yaml
```diff
@@ -50,7 +52,7 @@ serviceAccount:
 
 # Deployment configuration (fixed)
 replicaCount: 1
-namespace: default
+namespace: ""
```
paulgmiller, line 55: this lets us take namespace from helm command line. Should we remove it entirely and always depend on release.namespace?

helm/eviction-autoscaler/templates/_helpers.tpl
paulgmiller: https://github.com/Azure/eviction-autoscaler/pull/42/files#r2223942268

test/e2e/e2e_test.go
```diff
@@ -135,13 +135,35 @@ var _ = Describe("controller", Ordered, func() {
...
+			tag := imgParts[1]
+
+			// Use `helm upgrade --install` so that the test can be re-run without manual cleanup.
+			helmArgs := []string{
+				"upgrade", "--install", "eviction-autoscaler", "helm/eviction-autoscaler",
+				"--namespace", "kube-system", "--create-namespace",
+				"--set", fmt.Sprintf("image.repository=%s", repo),
+				"--set", fmt.Sprintf("image.tag=%s", tag),
+				"--set", "image.pullPolicy=IfNotPresent",
+				"--set", "serviceMonitor.enabled=false",
```
paulgmiller, line 156: lets rip this out.

test/e2e/e2e_test.go
```diff
@@ -135,13 +135,35 @@ var _ = Describe("controller", Ordered, func() {
...
+			repo := imgParts[0]
+			tag := imgParts[1]
+
+			// Use `helm upgrade --install` so that the test can be re-run without manual cleanup.
+			helmArgs := []string{
+				"upgrade", "--install", "eviction-autoscaler", "helm/eviction-autoscaler",
+				"--namespace", "kube-system", "--create-namespace",
+				"--set", fmt.Sprintf("image.repository=%s", repo),
+				"--set", fmt.Sprintf("image.tag=%s", tag),
+				"--set", "image.pullPolicy=IfNotPresent",
```
paulgmiller, line 155: why? Maybe because kind can't support always cause this isn't in an acr. Let us know though.

helm/eviction-autoscaler/values.yaml
```diff
@@ -39,6 +39,8 @@ controllerConfig:
   pdb:
     create: true
 
+
```
paulgmiller, line 42: repository: paulgmiller/k8s-pdb-autoscaler
  tag: latest
  
  Time to start pointing at dalec image in seperate pr.

test/e2e/e2e_test.go
```diff
@@ -56,7 +56,7 @@ func init() {
 	utilruntime.Must(types.AddToScheme(scheme))
 }
 
-// SYNC with kustomize file
+// Test namespace for eviction-autoscaler
```
paulgmiller, line 59: below you had code namespace though?

paulgmiller: Fin with either way always testing in kube-system or seperate namespace just need to be consistent.

paulgmiller (changes requested):
Mostly minor stuff to simplify down.

hack/install.sh
paulgmiller: or make it equivalent to helm upgrade --install you are doing in e2e.

helm/eviction-autoscaler/templates/servicemonitor.yaml
skuchipudi295: this change is okay to me

helm/eviction-autoscaler/templates/servicemonitor.yaml
```diff
@@ -15,11 +16,12 @@ spec:
   - path: /metrics
     port: http
     scheme: http
-    interval: 30s
-    scrapeTimeout: 10s
+    interval: {{ .Values.serviceMonitor.interval }}
+    scrapeTimeout: {{ .Values.serviceMonitor.scrapeTimeout }}
```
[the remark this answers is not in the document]
skuchipudi295: aren't the default values involves.yaml file?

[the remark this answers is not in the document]
skuchipudi295: aren't the default values involves.yaml file?

helm/eviction-autoscaler/values.yaml
skuchipudi295: no this should eventually point to acr, we can always have a test acr and make it point to that

helm/eviction-autoscaler/values.yaml
skuchipudi295: why eviction-autoscaler.ns?

test/e2e/e2e_test.go
skuchipudi295: lets use sepearet namespace?

helm/eviction-autoscaler/templates/deployment.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ include "eviction-autoscaler.fullname" . }}-controller-manager
-  namespace: {{ .Values.namespace }}
```
skuchipudi295, line 5: I actually dont understand why this is needed? values.yaml already has a default?

helm/eviction-autoscaler/templates/deployment.yaml
```diff
@@ -65,11 +63,3 @@ spec:
           allowPrivilegeEscalation: false
           capabilities:
             drop: ["ALL"]
-        volumeMounts:
```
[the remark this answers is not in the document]
skuchipudi295: okay to me..  but we will need config map though... whether to surge on all with exclude or surge on specific 

if we want use to configure, they should at least  tell us which annotation(exclude/include) to look for

helm/eviction-autoscaler/templates/namespace.yaml
skuchipudi295: yes please

helm/eviction-autoscaler/templates/deployment.yaml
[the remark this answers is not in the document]
Javier090: should I remove for now then add back if we decide to use a config map?

test/e2e/e2e_test.go
```diff
@@ -135,13 +135,35 @@ var _ = Describe("controller", Ordered, func() {
...
+			helmArgs := []string{
+				"upgrade", "--install", "eviction-autoscaler", "helm/eviction-autoscaler",
+				"--namespace", "kube-system", "--create-namespace",
+				"--set", fmt.Sprintf("image.repository=%s", repo),
+				"--set", fmt.Sprintf("image.tag=%s", tag),
+				"--set", "image.pullPolicy=IfNotPresent",
+				"--set", "serviceMonitor.enabled=false",
+			}
+
+			cmd = exec.Command("helm", helmArgs...)
```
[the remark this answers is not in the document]
Javier090: as a comment that it was test on "this version"

test/e2e/e2e_test.go
Javier090: my understanding is that its because the kind is local cache only so you need IfNotPresent since theres no remote registry https://kind.sigs.k8s.io/docs/user/quick-start/#loading-an-image-into-your-cluster

Makefile
```diff
@@ -114,34 +114,28 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
...
-	$(KUSTOMIZE) build config/crd | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
+.PHONY: helm-uninstall
+helm-uninstall: ## Uninstall the controller using Helm.
+	helm uninstall eviction-autoscaler --namespace kube-system
 
 .PHONY: deploy
-deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
-	cd config/manager && $(KUSTOMIZE) edit set image paulgmiller/k8s-pdb-autoscaler=${IMG}
-	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
+deploy: helm-install ## Deploy controller to the K8s cluster specified in ~/.kube/config.
```
paulgmiller, line 135: this was only different than install in that it took an img. either just remove or replicate that?

hack/install.sh
```diff
@@ -1,128 +1,27 @@
...
-# Apply Webhook svc
-#apply_yaml $WEBHOOK_SVC_FILE
-
-
-echo "Installation completed."
+# Simple installation script using Helm chart
+# This replaces the legacy manual installation with a Helm-based approach
+
+# Default values
+IMAGE_REPOSITORY="${IMAGE_REPOSITORY:-paulgmiller/k8s-pdb-autoscaler}"
```
paulgmiller, line 8: point at mcr/dalec image.

helm/eviction-autoscaler/values.yaml
paulgmiller: e me the default should always be the latest in mcr we can always override to point a our acr for tests or otherwise

paulgmiller (approved):

Javier090:
@microsoft-github-policy-service agree company="Microsoft"

PR #44, opened 2025-08-05
Labels

Javier090:
This PR adds logic for analyzing pod failure patterns based on pod age and integrates this into scaling decision metrics for the eviction autoscaler. The main focus is to improve observability and predictive scaling by tracking whether scaling up is likely to be effective, based on which pods (oldest, newest, or random). It also updates the Helm chart image configuration and makes a minor artifact naming change.

### Metrics and Scaling Logic Enhancements

* Added new metrics: `PodAgeFailurePatternGauge` and `ScalingEffectivenessGauge`, which track failure patterns by pod age and predict whether scaling up would be helpful, respectively. These metrics are registered and updated during reconciliation. 
* Enhanced the `ScalingOpportunityCounter` metric to include new labels: `age_pattern` and `likely_helpful`, providing richer context for scaling opportunities.
* Implemented pod age failure pattern analysis and scaling effectiveness prediction logic, with integration into the controller's reconciliation loop to update metrics and inform scaling decisions. 

### Helm Chart and Workflow Updates

* Updated the container image repository and tag in `helm/eviction-autoscaler/values.yaml` to use a new, versioned image from Microsoft's registry.
* Changed the artifact name in the GitHub workflow from `SARIF file` to `sarif-file` for consistency.


.github/workflows/scorecard.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       - name: "Upload artifact"
         uses: actions/upload-artifact@97a0fba1372883ab732affbe8f94b823f91727db # v3.pre.node20
         with:
-          name: SARIF file
+          name: sarif-file
```
paulgmiller, line 64: unrelated?

helm/eviction-autoscaler/values.yaml
```diff
@@ -4,8 +4,8 @@
 
 # Container image configuration
 image:
-  repository: paulgmiller/k8s-pdb-autoscaler
-  tag: latest
+  repository: mcr.microsoft.com/oss/v2/eviction-autoscaler/eviction-autoscaler
```
paulgmiller, line 7: @skuchipudi295  fyi.

internal/metrics/metrics.go
```diff
@@ -187,6 +227,172 @@ func GetScalingSignal(pdb *policyv1.PodDisruptionBudget) string {
 	return PDBBlockedSignal
 }
 
+// GetScalingOpportunityLabels gets age pattern and effectiveness labels for scaling opportunities
+func GetScalingOpportunityLabels(ctx context.Context, c client.Client, pdb *policyv1.PodDisruptionBudget) (string, string) {
```
paulgmiller, line 231: probabluy want a new file if not a new pacakge for AnalyzePodAgeFailurePattern and mabye GetScalingOpportunityLabels

internal/metrics/metrics.go
```diff
@@ -187,6 +227,172 @@ func GetScalingSignal(pdb *policyv1.PodDisruptionBudget) string {
...
+	pattern, err := AnalyzePodAgeFailurePattern(ctx, c, pdb)
+	if err != nil {
+		return "unknown", "unknown"
+	}
+	
+	isHelpful := PredictScalingEffectiveness(pattern)
+	helpfulLabel := UnlikelyHelpfulStr
+	if isHelpful {
```
paulgmiller, line 239: nit is isHelpful uncessary variable also PredictScalingEffectiveness can return a string instead a bool if this is the only place used.

internal/metrics/metrics.go
```diff
@@ -187,6 +227,172 @@ func GetScalingSignal(pdb *policyv1.PodDisruptionBudget) string {
...
+		if condition.Type == corev1.PodReady {
+			return condition.Status == corev1.ConditionTrue
+		}
+	}
+
+	return false
+}
+
+// PredictScalingEffectiveness predicts whether scaling up would be helpful based on failure pattern
+func PredictScalingEffectiveness(pattern string) bool {
```
paulgmiller, line 345: methods that aren't used outside package should be lower case. 
Be cautious about what you expose its your defactor api. Default to lower case and only expose what you need

internal/metrics/metrics.go
```diff
@@ -187,6 +227,172 @@ func GetScalingSignal(pdb *policyv1.PodDisruptionBudget) string {
...
+		// If oldest pods are failing, scaling up helps by giving new pods time to stabilize
+		return true
+	case NewestFailingPattern:
+		// If newest pods are failing, scaling up won't help much as new pods will likely also fail
+		return false
+	case RandomFailingPattern:
+		// Random failures - unclear if scaling helps default to maybe?
+		return false
+	case AllHealthyPattern:
+		// All pods healthy, no need to scale
```
paulgmiller, line 357: unless minavailable == replicas :)

internal/metrics/metrics.go
```diff
@@ -187,6 +227,172 @@ func GetScalingSignal(pdb *policyv1.PodDisruptionBudget) string {
...
+		return false
+	default:
+		return false
```
paulgmiller, line 360: should never happen? Panic or error

internal/controller/evictionautoscaler_controller.go
```diff
@@ -180,8 +188,8 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
 	//still at a scaled out state check if we can scale back down
 	if target.GetReplicas() > EvictionAutoScaler.Status.MinReplicas { //would we ever be below min replicas
 
-		// Track scaling opportunity
-		metrics.ScalingOpportunityCounter.WithLabelValues(EvictionAutoScaler.Namespace, EvictionAutoScaler.Spec.TargetName, metrics.ScaleDownAction, metrics.CooldownElapsedSignal).Inc()
+		// Track scaling opportunity (for scale down we don't need age pattern analysis)
+		metrics.ScalingOpportunityCounter.WithLabelValues(EvictionAutoScaler.Namespace, EvictionAutoScaler.Spec.TargetName, metrics.ScaleDownAction, metrics.CooldownElapsedSignal, "none", "true").Inc()
```
paulgmiller, line 192: Is this actually an opportunity? we already scaled up so wy keep incrementing?

internal/metrics/metrics.go
```diff
@@ -187,6 +227,172 @@ func GetScalingSignal(pdb *policyv1.PodDisruptionBudget) string {
...
+	err = c.List(ctx, &podList, &client.ListOptions{
+		Namespace:     pdb.Namespace,
+		LabelSelector: selector,
+	})
+	if err != nil {
+		return AllHealthyPattern, err
+	}
+
+	if len(podList.Items) == 0 {
+		return AllHealthyPattern, nil
```
paulgmiller, line 265: this isn't really all healthy this is a different pattern maybe (an abandoned pdb? we mention this elsewhere in the code)

internal/metrics/metrics.go
```diff
@@ -187,6 +227,172 @@ func GetScalingSignal(pdb *policyv1.PodDisruptionBudget) string {
...
+	// If all pods are healthy
+	if len(unhealthyPods) == 0 {
+		return AllHealthyPattern, nil
+	}
+
+	// Analyze failure pattern 
+	totalPods := len(podList.Items)
+	totalUnhealthyPods := len(unhealthyPods)
+	
+	// If less than 2 pods, can't determine pattern clearly
```
paulgmiller, line 290: handle this early with if len(podList.Items) == 0 

This means thay have o have a minavailable = 1 or they woudn't be blocked?

internal/metrics/metrics.go
```diff
@@ -179,12 +217,169 @@ func GetPDBCreatedByUsLabel(annotations map[string]string) string {
...
+	PodAgeFailurePatternGauge.WithLabelValues(pdb.Namespace, pdb.Name, AllHealthyPattern).Set(0)
+	PodAgeFailurePatternGauge.WithLabelValues(pdb.Namespace, pdb.Name, AbandonedPDBPattern).Set(0)
+	PodAgeFailurePatternGauge.WithLabelValues(pdb.Namespace, pdb.Name, UnknownPattern).Set(0)
+
+	// Set the current pattern
+	PodAgeFailurePatternGauge.WithLabelValues(pdb.Namespace, pdb.Name, pattern).Set(1)
+
+	// Update scaling opportunity counter with signal
+	signal := GetScalingSignal(ctx, c, pdb)
+	ScalingOpportunityCounter.WithLabelValues(
```
paulgmiller, line 377: it seems wierd to increment the opportunity counter here. If we've handled all the evictions at 
	if EvictionAutoScaler.Spec.LastEviction == EvictionAutoScaler.Status.LastEviction {
	
	Why are we saying theres anotehr scaling oppurtunt? 
	
	The guages are fine but they also only get updated when an eviction starts and is then handled so might want to clarify that about them.

internal/metrics/metrics.go
```diff
@@ -179,12 +217,169 @@ func GetPDBCreatedByUsLabel(annotations map[string]string) string {
...
+	// First check if minAvailable == replicas (highest priority signal)
+	if pdb.Spec.MinAvailable != nil && pdb.Status.CurrentHealthy > 0 {
+		minAvailable := pdb.Spec.MinAvailable.IntValue()
+		if minAvailable == int(pdb.Status.CurrentHealthy) {
+			return MinAvailableEqualsReplicasSignal
+		}
+	}
+
+	// Then check pod failure patterns
+	pattern, err := analyzePodAgeFailurePattern(ctx, c, pdb)
```
paulgmiller, line 230: should his just also return a signal reather than a pattern?

internal/metrics/metrics.go
```diff
@@ -179,12 +217,169 @@ func GetPDBCreatedByUsLabel(annotations map[string]string) string {
...
+		if !isPodHealthy(&pod) {
+			if i < oldestHalf {
+				oldestFailureCount++
+			} else {
+				newestFailureCount++
+			}
+		}
+	}
+
+	// Follow majority logic
```
paulgmiller, line 323: I am not sure if half and 66% is what we want in the end.. I think I'm fine with it for his iteration but its not theres anything special about half. 
Another way of doing this would be to use the time from the eviction notice which we have. If pods that are crashing are older than the last eviction that'd be a good clue but we keep updating the eviction after cooldown.  Coud go back to keeping a list of evictions.

A great place to leave your thoughts in the comments.

paulgmiller: Could add to status start of unhdnaled evictions

Javier090: I think think we could keep compariing cration and eviction times is that we have to keep updating the  timestamp after cooldown? also making timing analysis kin of hard to trakc since its only tracked after cooldown
A list is probably the cleanest? it would be good to compare pod age vs when evictions started occurring?

Javier090: eviciton timing one is prob best, I think thats where we should pivot to

PR #58, opened 2025-10-15
using dalec instead of ko

skuchipudi295:
using dalec instead of ko; because of I am fed up with ko

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -11,8 +11,20 @@ RELEASE_ACR_FQDN="${RELEASE_ACR}.azurecr.io"
 IMAGE_REPO="${RELEASE_ACR_FQDN}/public/aks/eviction-autoscaler"
 repo_path="public/aks/eviction-autoscaler"  # adjust if your ko publish path changes
 
-latest_tag=$(az acr repository show-tags -n "$RELEASE_ACR" --repository "$repo_path" -o tsv | \
+set +e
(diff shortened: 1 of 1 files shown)


hack/release.sh
```diff
@@ -9,11 +9,22 @@ commit_sha="$(git rev-parse HEAD)"
 RELEASE_ACR="${RELEASE_ACR:-aksmcrimagescommon}"
 RELEASE_ACR_FQDN="${RELEASE_ACR}.azurecr.io"
 IMAGE_REPO="${RELEASE_ACR_FQDN}/public/aks/eviction-autoscaler"
-repo_path="public/aks/eviction-autoscaler/cmd"  # adjust if your ko publish path changes
+repo_path="public/aks/eviction-autoscaler"  # adjust if your ko publish path changes
 
-# List all tags, filter for semver, sort, and get the latest
-latest_tag=$(az acr repository show-tags -n "$RELEASE_ACR" --repository "$repo_path" -o tsv | \
+set +e
```
paulgmiller: bash is  not my forte but seems wierd to have a set +e nto at top of script is only latest tag exported?

skuchipudi295: using set +e just for the az acr repository show-tags command is the right approach, so you can catch errors from that command only, while keeping the rest of your script strict (set -euo pipefail).

Only latest_tag and acr_err are exported from that block.
The rest of your script will still exit on errors, which is best practice.

PR #70, opened 2025-11-17
add helm lint and helm template render to ci.yaml

skuchipudi295:
Adds helm-lint and helm-template targets to the Makefile for local development and CI use
Integrates these Helm validation steps into the CI workflow to run on every push and pull request


paulgmiller (approved):
hooray linters

PR #101, opened 2026-04-17
Sync PDB minAvailable with HPA/KEDA autoscaler floor

skuchipudi295:
- Add AutoscalerToPDBReconciler to watch HPA/ScaledObject changes and update PDB minAvailable
- Add autoscaler_helpers.go with shared functions: findScaledObjectForTarget, findHPAForTarget, HasAutoscaler, ResolveMinReplicas
- Update DeploymentToPDBReconciler to skip PDB updates when HPA/KEDA controls the deployment
- Update CreatePDBForDeployment to use ResolveMinReplicas for initial PDB creation
- Add read-only RBAC for horizontalpodautoscalers and scaledobjects
- Add unit tests for PDB minAvailable with HPA
- Add e2e test for PDB minAvailable from HPA minReplicas


internal/controller/deployment_to_pdb_controller.go
```diff
@@ -140,33 +140,51 @@ func (r *DeploymentToPDBReconciler) updateMinAvailableAsNecessary(ctx context.Co
...
-		pdb.Spec.MinAvailable = &intstr.IntOrString{IntVal: *deployment.Spec.Replicas}
-		err := r.Update(ctx, &pdb)
+	// No autoscaler — only proceed if the deployment generation actually changed
+	if EvictionAutoScaler.Status.TargetGeneration == deployment.GetGeneration() {
+		return nil
+	}
+
+	// Track deployment.spec.replicas directly.
+	// But skip if the replica change was caused by our own eviction surge.
+	if surgeReplicas, exists := deployment.Annotations[EvictionSurgeReplicasAnnotationKey]; exists {
```
skuchipudi295, line 163: add a comment here 
//EvictionAutoScaler can fail between updating deployment and EvictionAutoScaler targetGeneration;
		//hence we need to rely on checking if annotation exists and compare with deployment.Spec.Replicas
		// no surge happened but customer already increased deployment replicas, then annotation would not exist

cmd/main.go
```diff
@@ -192,6 +192,16 @@ func main() {
 			os.Exit(1)
 		}
 		setupLog.Info("DeploymentToPDBReconciler setup completed")
+
+		if err = (&controllers.AutoscalerToPDBReconciler{
```
paulgmiller, line 198: assume I'll find out later but is this just HPA or also keda? Not clear from main

internal/controller/autoscaler_helpers.go
```diff
@@ -0,0 +1,154 @@
...
+
+	if err := c.List(ctx, list, client.InNamespace(namespace)); err != nil {
+		if meta.IsNoMatchError(err) {
+			return nil, errNotFound
+		}
+		return nil, err
+	}
+
+	for i := range list.Items {
+		item := &list.Items[i]
```
paulgmiller: are we mutating item? why not _, item := range list.Item

internal/controller/autoscaler_helpers.go
```diff
@@ -0,0 +1,154 @@
...
+		scaleTargetRef, found, err := unstructured.NestedMap(item.Object, "spec", "scaleTargetRef")
+		if err != nil || !found {
+			continue
+		}
+		name, _ := scaleTargetRef["name"].(string)
+		kind, _ := scaleTargetRef["kind"].(string)
+		if kind == "" {
+			kind = "Deployment" // KEDA default
```
paulgmiller, line 52: safer to just error?

internal/controller/autoscaler_helpers.go
```diff
@@ -0,0 +1,154 @@
...
+	if err := c.List(ctx, list, client.InNamespace(namespace)); err != nil {
+		if meta.IsNoMatchError(err) {
+			return nil, errNotFound
+		}
+		return nil, err
+	}
+
+	for i := range list.Items {
+		item := &list.Items[i]
+		scaleTargetRef, found, err := unstructured.NestedMap(item.Object, "spec", "scaleTargetRef")
```
paulgmiller, line 40: nit just pull in keda's crds an register them with your client? or leave a comemment sayign why not

internal/controller/autoscaler_helpers.go
```diff
@@ -0,0 +1,154 @@
...
+		if err != nil || !found {
```
paulgmiller: wotrht logging not found?

internal/controller/autoscaler_helpers.go
```diff
@@ -0,0 +1,154 @@
...
+		if name == targetName && strings.EqualFold(kind, targetKind) {
+			return item, nil
+		}
+	}
+
+	return nil, errNotFound
+}
+
+// findHPAForTarget looks for a standalone (non-KEDA-managed) HPA targeting the given workload.
+// KEDA-managed HPAs are filtered out because they are owned by the ScaledObject and should
```
paulgmiller, line 63: I learned soemthing new about keda :). Good comemnt!

internal/controller/autoscaler_helpers.go
paulgmiller: the hpa one below is so much "tighter"

internal/controller/autoscaler_helpers.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}
+	return false
+}
+
+// HasAutoscaler returns true if an HPA or KEDA ScaledObject targets this workload.
+// Returns an error on real API failures (not errNotFound) so the caller can retry.
+func HasAutoscaler(ctx context.Context, c client.Client, namespace, targetName, targetKind string) (bool, error) {
+	_, err := findScaledObjectForTarget(ctx, c, namespace, targetName, targetKind)
+	if err != nil && !errors.Is(err, errNotFound) {
```
paulgmiller: hypernit. you might make this clerer by inveting
if err == nil { 
     return true, nil
} 
if !errors.Is(err, errNotfoudn) {
   reutnr false ,err 
}

...rest

internal/controller/autoscaler_to_pdb_controller.go
```diff
@@ -0,0 +1,200 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// AutoscalerToPDBReconciler watches HPA and KEDA ScaledObject changes
```
paulgmiller, line 24: Say why not two controllers. Okay if you considered that and found it too verbose but noramlly a controller watches one type so leave us a clue why you made this decision

cmd/main.go
paulgmiller: https://github.com/Azure/eviction-autoscaler/pull/101/changes#r3120647822

internal/controller/autoscaler_to_pdb_controller.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// SetupWithManager registers watches on HPA and KEDA ScaledObject resources.
+func (r *AutoscalerToPDBReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	builder := ctrl.NewControllerManagedBy(mgr).
+		Named("autoscaler-to-pdb").
+		Watches(&autoscalingv2.HorizontalPodAutoscaler{},
+			handler.EnqueueRequestsFromMapFunc(requeueDeploymentFromHPA()))
+
+	// Try to watch KEDA ScaledObjects (only if the CRD is installed)
```
paulgmiller: if the customer installs keda do we notice and start watching or do we need a restart before we notice? either way uis fine just need to docuemtn.

paulgmiller: assume we need a restart.

internal/controller/autoscaler_to_pdb_controller.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	if !isEnabled {
+		return reconcile.Result{}, nil
+	}
+
+	// Resolve the target deployment from the HPA/ScaledObject's scaleTargetRef.
+	// The watch mappers translate HPA/ScaledObject events into deployment namespace/name keys.
+	// If the deployment doesn't exist (e.g. HPA outlived it), there's nothing to do.
+	var deployment v1.Deployment
+	if err := r.Get(ctx, req.NamespacedName, &deployment); err != nil {
```
paulgmiller: I'm confused I thought we are watching autoscalers. Why would req.namespacedName point at deployments.

internal/controller/autoscaler_to_pdb_controller.go
```diff
@@ -0,0 +1,200 @@
...
+	if _, surgeActive := deployment.Annotations[EvictionSurgeReplicasAnnotationKey]; surgeActive {
+		logger.V(1).Info("Surge active on deployment, skipping PDB minAvailable update",
+			"deployment", deployment.Name)
+		return reconcile.Result{}, nil
+	}
+
+	// Resolve the autoscaler's minimum replica floor (KEDA minReplicaCount > HPA minReplicas).
+	// If no autoscaler targets this deployment, the deployment controller owns PDB updates
+	// and we bail out. We pass 0 as the fallback (unused when found==true).
+	minAvailable, found, err := ResolveMinReplicas(ctx, r.Client, req.Namespace, req.Name, ResourceTypeDeployment, 0)
```
paulgmiller: AS with https://github.com/Azure/eviction-autoscaler/pull/101/changes#r3120670339

req is an auiotscaler not a deplopyemnt

internal/controller/autoscaler_to_pdb_controller.go
paulgmiller: Ah this is because of 	handler.EnqueueRequestsFromMapFunc(requeueDeploymentFromHPA()))?

Why do that requeue. Why not just get the autoscaler here then look up its deplpoyment.

internal/controller/deployment_to_pdb_controller.go
```diff
@@ -140,33 +140,51 @@ func (r *DeploymentToPDBReconciler) updateMinAvailableAsNecessary(ctx context.Co
...
-			newReplicas, err := strconv.Atoi(surgeReplicas)
-			if err != nil {
-				logger.Error(err, "unable to parse surge replicas from annotation NOT updating",
-					"namespace", deployment.Namespace, "name", deployment.Name, "replicas", surgeReplicas)
-				return err
-			}
+	// When HPA/KEDA targets this deployment, a separate controller (AutoscalerToPDBReconciler)
+	// is responsible for tracking their minReplicas/minReplicaCount and updating PDB minAvailable.
+	// This controller should not interfere with autoscaler-driven replica changes.
+	hasAS, err := HasAutoscaler(ctx, r.Client, deployment.Namespace, deployment.Name, ResourceTypeDeployment)
```
paulgmiller, line 150: could make this cheaper if you annoate the deploument but guess then have to unannotate it when deleted. Guess we're cached so don't need to worry about expense.

internal/controller/autoscaler_to_pdb_controller.go
```diff
@@ -0,0 +1,200 @@
...
+
+	// Find the PDB that matches this deployment's pod selector labels.
+	// Only consider PDBs created by this controller (ownedBy annotation) — user-managed PDBs
+	// should not be modified. If no controller-owned PDB exists, there's nothing to update.
+	pdb, found, err := findPDBForDeployment(ctx, r.Client, &deployment, true)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if !found {
+		return reconcile.Result{}, nil
```
paulgmiller, line 87: Why not create it here?  In my mental model this controller should do create and updates for hpa/keda owned deployments. Deployments controller should just bale early.

internal/controller/pdb_helpers.go
```diff
@@ -89,6 +89,17 @@ func CreatePDBForDeployment(ctx context.Context, c client.Client, deployment *v1
 	controller := true
 	blockOwnerDeletion := true
 
+	// Use KEDA/HPA minReplicas when available instead of deployment.spec.replicas,
+	// since the autoscaler controls the actual replica count and may have scaled above its floor.
+	var deployReplicas int32 = 1
+	if deployment.Spec.Replicas != nil {
+		deployReplicas = *deployment.Spec.Replicas
+	}
+	minAvailable, _, err := ResolveMinReplicas(ctx, c, deployment.Namespace, deployment.Name, ResourceTypeDeployment, deployReplicas)
```
paulgmiller, line 98: could pass in min replicas as an optional param when created from autoscaler? https://github.com/Azure/eviction-autoscaler/pull/101/changes#r3120695879

internal/controller/deployment_to_pdb_controller.go
```diff
@@ -140,33 +140,51 @@ func (r *DeploymentToPDBReconciler) updateMinAvailableAsNecessary(ctx context.Co
...
+	if err != nil {
+		return err
+	}
+	if hasAS {
+		logger.V(1).Info("HPA/KEDA controls this deployment, skipping PDB minAvailable update",
```
paulgmiller, line 155: are we skipping cration too? wasn't clear. I think we should but then autocaler controller should create.

paulgmiller (approved):
Approving but 3 big meta comments.
why not two seperate controllers one scaled object
why re-enqueu the deployment why not let reconcile recieve the autoscaler then just fetch the deployment. Then you don't have to re-lookup the autoscaler. 
why doesnt' the autoscaler controller create the pdb

internal/controller/autoscaler_helpers.go
skuchipudi295: KEDA's module pulls in 80+ direct dependencies (AWS, Azure, GCP SDKs, Kafka, etc.)
// which would massively bloat our dependency tree for just two types.

it will increase build time as well

internal/controller/autoscaler_to_pdb_controller.go
skuchipudi295: unless until customer doesnt create a deployment that is being targeted by keda, should we care?

skuchipudi295: yea we need restart

internal/controller/autoscaler_to_pdb_controller.go
skuchipudi295: A single controller handles both HPA and ScaledObject because the reconcile logic
 is identical (resolve the target deployment from scaleTargetRef → resolve min replicas
 floor → update PDB). Two separate controllers would duplicate this logic without any benefit.

internal/controller/autoscaler_to_pdb_controller.go
skuchipudi295: PDB creation is gated by the [pdb-create](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) annotation on the Deployment — the deployment controller is the natural place to check that
[CreatePDBForDeployment](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) already calls [ResolveMinReplicas](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — so even when the deployment controller creates the PDB for an HPA-managed deployment, it picks the correct initial [minAvailable](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) from the autoscaler floor
After creation, the autoscaler controller takes over [minAvailable](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) updates

PR #104, opened 2026-04-27
Surge strategy refactor

skuchipudi295:
This pull request introduces a new abstraction for surge scaling strategies in the eviction autoscaler controller, making the surge logic extensible and more robust. The main change is the introduction of the `SurgeApplier` interface and its implementations, which decouple the surge logic from the controller and allow for future support of multiple scaling mechanisms (such as KEDA and HPA). The controller is refactored to use this new abstraction, and comprehensive unit tests are added for the new logic.

**Surge strategy abstraction and integration:**

* Introduced the `SurgeApplier` interface and its implementations (`DeploymentSurgeApplier`, `CompositeSurgeApplier`) in `surge_strategy.go`, abstracting how surge scaling is applied and reverted for deployments. This paves the way for supporting multiple surge strategies (e.g., KEDA, HPA, and plain deployments).
* Refactored the main controller logic in `evictionautoscaler_controller.go` to use the new `SurgeApplier` abstraction for applying and reverting surges, replacing direct manipulation of deployment replicas and annotations. This makes the code more maintainable and extensible. [[1]](diffhunk://#diff-6c377cb9f9f1e6722559d6e45aa09d2475f9ae85c1b3db39debff112383b69ddR117-R141) [[2]](diffhunk://#diff-6c377cb9f9f1e6722559d6e45aa09d2475f9ae85c1b3db39debff112383b69ddL148-R166) [[3]](diffhunk://#diff-6c377cb9f9f1e6722559d6e45aa09d2475f9ae85c1b3db39debff112383b69ddL158-R186) [[4]](diffhunk://#diff-6c377cb9f9f1e6722559d6e45aa09d2475f9ae85c1b3db39debff112383b69ddL195-R213) [[5]](diffhunk://#diff-6c377cb9f9f1e6722559d6e45aa09d2475f9ae85c1b3db39debff112383b69ddL212-R222)

**Testing and reliability improvements:**

* Added comprehensive unit tests for `DeploymentSurgeApplier`, `CompositeSurgeApplier`, and the annotation helper functions in the new `surge_strategy_test.go` file, ensuring correct behavior of surge application, reversion, and annotation logic.

These changes significantly improve the flexibility and testability of the surge scaling logic, and lay the groundwork for supporting additional autoscaling strategies in the future.


internal/controller/surge_strategy.go
```diff
@@ -0,0 +1,157 @@
...
+	// Name returns a human-readable name for logging
+	Name() string
+}
+
+// detectSurgeApplier determines which surge strategy to use by building a
+// composite of all applicable appliers. DeploymentSurgeApplier is always
+// included as the base (it sets deployment.spec.replicas directly). KEDA and
+// HPA appliers are added when their resources target this workload, ensuring
+// their floors are raised before the deployment scales.
+func detectSurgeApplier(ctx context.Context, c client.Client, namespace, targetName, targetKind string, target Surger) (SurgeApplier, error) {
```
paulgmiller, line 38: do we do any targetkinds besides deployment? Shoul we just  not expose kind or error if something besides deployment if so?

internal/controller/surge_strategy.go
```diff
@@ -0,0 +1,157 @@
...
+
+	var appliers []SurgeApplier
+
+	// Check for KEDA ScaledObject targeting this workload
+	scaledObj, err := findScaledObjectForTarget(ctx, c, namespace, targetName, targetKind)
+	if err != nil && !errors.Is(err, errNotFound) {
+		return nil, fmt.Errorf("checking for KEDA ScaledObject: %w", err)
+	}
+	if scaledObj != nil {
+		logger.Info("Found KEDA ScaledObject for target, skipping deployment surge (KEDA strategy not yet implemented)",
```
paulgmiller: not actually skipping because appliers will still be zero and we'll ad deployment surge on line 62?

paulgmiller: Could return a no-op appkier that just does nothing in this case.

internal/controller/surge_strategy.go
```diff
@@ -0,0 +1,157 @@
...
+	return c.Update(ctx, d.target.Obj())
+}
+
+func (d *DeploymentSurgeApplier) Name() string {
+	return "deployment"
+}
+
+// --- CompositeSurgeApplier ---
+// Surges multiple resources (e.g., both KEDA ScaledObject and standalone HPA) when both
+// target the same workload. This prevents the standalone HPA from blocking scale-up when
```
paulgmiller, line 152: doesn't matter for this iteration since they're not implmented but does the scaled objet not won the hpa when it exits? would it not scale up the hpa rather than us? Mostly my infamilarity with keda here.

internal/controller/surge_strategy.go
```diff
@@ -0,0 +1,157 @@
...
+// SurgeApplier abstracts the mechanism for temporarily increasing minimum replicas.
+// Depending on whether KEDA, HPA, or neither is present, a different implementation is used.
+//
+// For multi-resource strategies (HPA, KEDA): the HPA/KEDA floor is raised first, then
+// deployment replicas are set directly for immediate effect. On failure, the reconcile
+// loop retries ApplySurge idempotently until the deployment write succeeds.
+type SurgeApplier interface {
+	// ApplySurge sets the minimum replica count to surgeReplicas.
+	// Callers may invoke this multiple times; implementations must be idempotent.
+	ApplySurge(ctx context.Context, c client.Client, surgeReplicas int32) error
```
paulgmiller: you could remove c client.Client from the apply/revert methods and just give the client to the struct implmenting it so it can apply/revert itself. Smaller interfacs usually nicer but might be tyring to be too clever.

Also should the applier also kbe in charge of knowing how much to surge and revert or should an outside actor drive that?

internal/controller/surge_strategy.go
```diff
@@ -0,0 +1,157 @@
...
+
+type CompositeSurgeApplier struct {
+	appliers []SurgeApplier
+	target   Surger
+}
+
+var _ SurgeApplier = &CompositeSurgeApplier{}
+
+func (comp *CompositeSurgeApplier) ApplySurge(ctx context.Context, c client.Client, surgeReplicas int32) error {
+	for _, applier := range comp.appliers {
```
paulgmiller, line 176: will ordering be importiant? if so should revets happen in opposite order?

internal/controller/evictionautoscaler_controller.go
```diff
@@ -114,13 +114,31 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 	// Check if the resource version has changed or if it's empty (initial state)
 	if EvictionAutoScaler.Status.TargetGeneration == 0 || EvictionAutoScaler.Status.TargetGeneration != target.Obj().GetGeneration() {
-		logger.Info("Target resource version changed resetting min replicas", "kind", EvictionAutoScaler.Spec.TargetKind, "targetname", EvictionAutoScaler.Spec.TargetName, "currentGeneration", target.Obj().GetGeneration(), "previousGeneration", EvictionAutoScaler.Status.TargetGeneration)
-		// The resource version has changed, which means someone else has modified the Target.
-		// To avoid conflicts, we update our status to reflect the new state and avoid making further changes.
 		EvictionAutoScaler.Status.TargetGeneration = target.Obj().GetGeneration()
-		EvictionAutoScaler.Status.MinReplicas = target.GetReplicas()
+		// Don't reset MinReplicas if a surge is in progress (e.g., HPA/KEDA-driven scaling
+		// changes the deployment generation as part of the surge, not a user change).
+		if hasTargetAnnotation(target) {
```
paulgmiller: Should this be a method on the surge applier?

paulgmiller (approved):
This looks good though probably want to actually hpa/keda skip till you implment. (right now it falls back)

internal/controller/surge_strategy.go
skuchipudi295: there can be an hpa controlling deployment and customer deployed keda & never deleted old hpa... so we got do both.. as per my discussion with @Alex leites

internal/controller/surge_strategy.go
skuchipudi295: controller decides what (amount, timing, cooldown) → applier decides how (which K8s resources to mutate)

internal/controller/surge_strategy.go
skuchipudi295: still want to find hpa if no keda In this order

internal/controller/surge_strategy.go
skuchipudi295: Apply order: KEDA → HPA → Deployment — floor must be raised before replicas, or the autoscaler races and scales back down.
Revert order: Same (not reversed) — floor must be lowered before replicas, or the autoscaler immediately scales back up to the still-surged floor.

internal/controller/evictionautoscaler_controller.go
skuchipudi295: yes!!

PR #108, opened 2026-05-04
Implement KEDA surge strategy

skuchipudi295:
This pull request adds first-class support for KEDA ScaledObject surge handling, improves observability and documentation for surge operations, and introduces comprehensive unit tests for surge logic. The main features are the implementation of the `KEDASurgeApplier`, a clear annotation strategy for surge state tracking, and improved logging for HPA and KEDA surge transitions.

**KEDA Surge Support:**

* Added a new `KEDASurgeApplier` implementation (`internal/controller/keda_surge_applier.go`) that surges by raising the KEDA ScaledObject's `minReplicaCount` and annotating it, while also setting deployment replicas directly for immediate effect. Handles both applying and reverting surges, with safe and idempotent logic.
* Updated the surge detection logic to use the new `KEDASurgeApplier` when a KEDA ScaledObject is present, instead of skipping surge.

**Surge State Tracking and Documentation:**

* Updated the `README.md` to document the new KEDA-aware surge behavior and clearly describe the annotation strategy for both HPA and KEDA surges, including how to inspect surge state. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R31) [[2]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R625-R649)

**Testing and Observability:**

* Added comprehensive unit tests for both `HPASurgeApplier` and `KEDASurgeApplier`, covering surge application, idempotency, reversion, and state reporting.
* Improved logging for HPA and KEDA surge and revert operations for better traceability in controller logs. [[1]](diffhunk://#diff-46b77c9482e8d1c15132933e168e7db288e28bf623213576bd903bf8f7ed763bR57-R60) [[2]](diffhunk://#diff-46b77c9482e8d1c15132933e168e7db288e28bf623213576bd903bf8f7ed763bL116-R123)

These changes ensure surge logic is robust, observable, and well-documented for both HPA and KEDA-based autoscaling scenarios.


test/e2e/e2e_test.go
```diff
@@ -1244,5 +1248,240 @@ var _ = Describe("controller", Ordered, func() {
 			cmd = exec.Command("kubectl", "delete", "namespace", testNs)
 			_, _ = utils.Run(cmd)
 		})
+
+		// Test 4: KEDA surge strategy - when a ScaledObject exists, surge by updating minReplicaCount
+		It("should surge via KEDA ScaledObject minReplicaCount when a ScaledObject targets the deployment", func() {
```
tallaxes, line 1253: This covers the happy path, but would be stronger with some extra checks: original-min-replicas is set/removed, the Deployment is not annotated, and maybe PDB minAvailable tracking?

config/rbac/role.yaml
```diff
@@ -73,6 +73,7 @@ rules:
   verbs:
   - get
   - list
+  - update
```
tallaxes, line 76: Added here, but kubebuilder RBAC marker for ScaledObject `update` seems to be missing (either in autoscaler_to_pbd_controller.go, or elsewhere?); and this gets regenerated on make test/make manifests ...

tallaxes (commented):
Overall LGTM; left some minor questions
 
Do we intend to support KEDA + standalone HPA (same target) case? If not - let's document (and maybe detect?) this. If yes (which some of the comments appear to imply) the semantics may need more hardening (e.g. currently ResolveMinReplicas uses one baseline and prioritizes KEDA over HPA?) and test coverage ...

skuchipudi295:
we do support keda + standalone hpa --> same target today

skuchipudi295:
> Overall LGTM; left some minor questions
> 
> Do we intend to support KEDA + standalone HPA (same target) case? If not - let's document (and maybe detect?) this. If yes (which some of the comments appear to imply) the semantics may need more hardening (e.g. currently ResolveMinReplicas uses one baseline and prioritizes KEDA over HPA?) and test coverage ...

we do support support KEDA + standalone HPA (same target) case

skuchipudi295:
> > Overall LGTM; left some minor questions
> > Do we intend to support KEDA + standalone HPA (same target) case? If not - let's document (and maybe detect?) this. If yes (which some of the comments appear to imply) the semantics may need more hardening (e.g. currently ResolveMinReplicas uses one baseline and prioritizes KEDA over HPA?) and test coverage ...
> 
> we do support support KEDA + standalone HPA (same target) case

keda always takes precedence over standalone hpa while calculating surge

internal/controller/keda_surge_applier_test.go
```diff
@@ -0,0 +1,296 @@
...
+	Describe("ApplySurge annotations (in-memory)", func() {
+		// Note: Full ApplySurge with c.Update requires KEDA CRD installed in envtest.
+		// These tests verify the annotation and minReplicaCount logic in isolation.
+
+		It("should read originalMin from existing minReplicaCount", func() {
+			obj := createScaledObject("test-so", "default", "test-deploy", 2, 10)
+			val, found, _ := unstructured.NestedInt64(obj.Object, "spec", "minReplicaCount")
+			Expect(found).To(BeTrue())
+			Expect(val).To(Equal(int64(2)))
+		})
```
tallaxes: From here down, nothing tests KEDASurgeApplier functionality, it basically just exercises Kubernetes apimachinery ...

tallaxes:
> > > Do we intend to support KEDA + standalone HPA (same target) case? If not - let's document (and maybe detect?) this. If yes (which some of the comments appear to imply) the semantics may need more hardening (e.g. currently ResolveMinReplicas uses one baseline and prioritizes KEDA over HPA?) and test coverage ...
> > 
> > 
> > we do support support KEDA + standalone HPA (same target) case

Can we rethink this? I am thinking we should treat ScaledObject + standalone HPA on the same target as unsupported. Eviction Autoscaler can't really fix the issue with multiple writers, and I don't think it should try. KEDA already creates/owns its HPA, and validates against an unmanaged HPA on the same scale target unless ownership is explicitly transferred or validation is bypassed. Given that, detectSurgeApplier should either use the KEDA strategy only, or detect the extra standalone HPA and surface a clear unsupported configuration. This would also remove the need for the HPA “don’t lower floor” guard and avoid the current revert path lowering a standalone HPA that was skipped during apply.

test/e2e/e2e_test.go
tallaxes: This is not addressed yet, this needs to go into E2E tests; tests added to keda_surge_applier_tests.go in https://github.com/Azure/eviction-autoscaler/pull/108/commits/c079466829852b9de4e5ae47e372d4161e06eb4f  do not test real behavior

skuchipudi295:
> > > > Do we intend to support KEDA + standalone HPA (same target) case? If not - let's document (and maybe detect?) this. If yes (which some of the comments appear to imply) the semantics may need more hardening (e.g. currently ResolveMinReplicas uses one baseline and prioritizes KEDA over HPA?) and test coverage ...
> > > 
> > > 
> > > we do support support KEDA + standalone HPA (same target) case
> 
> Can we rethink this? I am thinking we should treat ScaledObject + standalone HPA on the same target as unsupported. Eviction Autoscaler can't really fix the issue with multiple writers, and I don't think it should try. KEDA already creates/owns its HPA, and validates against an unmanaged HPA on the same scale target unless ownership is explicitly transferred or validation is bypassed. Given that, detectSurgeApplier should either use the KEDA strategy only, or detect the extra standalone HPA and surface a clear unsupported configuration. This would also remove the need for the HPA “don’t lower floor” guard and avoid the current revert path lowering a standalone HPA that was skipped during apply.

makes sense actually

internal/controller/keda_surge_applier.go
```diff
@@ -0,0 +1,184 @@
...
+// KEDA/HPA from scaling back down below the surge value.
+//
+// Note: the HPA controller (which KEDA manages) uses the /scale subresource to set
+// deployment replicas, which is equivalent to updating deployment.spec.replicas
+// directly. The /scale subresource does NOT bump deployment metadata.generation,
+// while a full Update does (if spec changes).
+
+type KEDASurgeApplier struct {
+	client       client.Client
+	scaledObject *unstructured.Unstructured
```
paulgmiller: Still seems like you could load keda objectsd and use real types.

paulgmiller: https://chatgpt.com/share/69fcf1b4-3678-83e8-9d06-92cfd2285c25

paulgmiller: https://github.com/Azure/eviction-autoscaler/blob/f7c071e88fddbfa30ab89ab477c8e6612a6b20fa/cmd/main.go#L52-L54 add it to scheme here?

paulgmiller (approved):
Really enourage you to use structured rather than unstructured client.

skuchipudi295:
> Really enourage you to use structured rather than unstructured client.

done

tallaxes (commented):
Thanks for the significant update here, overall direction is much cleaner now, and E2E coverage is quite a bit better. Some additional findings - agent-assisted, but I did verify each one, and they all make sense to me:

#### Blocking / correctness

1. `ResolveMinReplicas` mishandles omitted KEDA `minReplicaCount`

KEDA defaults an omitted `spec.minReplicaCount` to `0`, but `ResolveMinReplicas` only returns the KEDA floor when `scaledObj.Spec.MinReplicaCount != nil`. If the field is omitted, the code falls through to standalone HPA / deployment replicas and can treat a scale-to-zero KEDA workload as having a higher minimum than it actually does.

This matters because the PDB floor and EA baseline should reflect KEDA's effective min replica count. Please treat a matching ScaledObject with nil `minReplicaCount` as found with min `0`, rather than falling through.

2. Original-min annotation can be overwritten if the surge value changes while a surge is active

`KEDASurgeApplier.ApplySurge` only skips the ScaledObject update when the existing `evictionSurgeReplicas` annotation already equals the new surge value. If an active surge is first applied from `1 -> 2`, then a later reconcile applies surge `3`, the method recalculates `originalMin` from the already-surged `spec.minReplicaCount=2` and overwrites `OriginalMinReplicasAnnotationKey`. `RevertSurge` would then restore `2` instead of the true original `1`.

Please preserve the existing original-min annotation once it has been set, and only initialize it when absent. The same pattern appears to exist in the HPA applier as well, so it is worth fixing both.

#### Medium

3. Unsupported KEDA + standalone HPA is returned as a retrying reconcile error

`detectSurgeApplier` now rejects a ScaledObject and standalone HPA targeting the same deployment, which I agree with. However, the caller returns that error directly from reconcile. Since this is a deterministic unsupported configuration, the controller will keep retrying and users only see the problem in logs.

Please surface this on `EvictionAutoScaler.Status` instead, for example `Degraded=True` with reason `UnsupportedAutoscalerConfiguration`, and return without continuous requeue churn. I would avoid passing a normal `NoOpSurgeApplier` through the existing scale-up path unless that path is changed not to record a successful scale action.

#### Low priority cleanup

4. Appliers update cached fields before the API write succeeds

`KEDASurgeApplier` assigns `k.scaledObject = obj` before `client.Update` succeeds. Today this is mostly harmless because appliers are rebuilt per reconcile and the method returns on error, but it makes the in-memory state look committed when the API write failed. Please move the assignment after the successful update or remove it entirely. The same applies to revert and to the HPA applier.

5. `CompositeSurgeApplier` and `NoOpSurgeApplier` are now dead code

`detectSurgeApplier` only returns KEDA, HPA, or deployment appliers. The composite path was useful when the implementation attempted to combine multiple writers, but this PR now makes strategies mutually exclusive and rejects KEDA + standalone HPA. Please remove the unused appliers and the corresponding composite tests rather than leaving dead code plus a future-cleanup comment.

#### Testing gaps

Please add focused unit coverage for the edge cases above:

- KEDA `ScaledObject` with omitted `minReplicaCount` resolves as autoscaler found with min `0`.
- Reapplying KEDA surge with a different surge value preserves the original-min annotation and reverts to the true original floor. (And the equivalent HPA changed-surge-value case, if the HPA applier has the same overwrite behavior.)
- ScaledObject + standalone HPA on the same target produces the intended non-retry / degraded status behavior.

skuchipudi295:
> Thanks for the significant update here, overall direction is much cleaner now, and E2E coverage is quite a bit better. Some additional findings - agent-assisted, but I did verify each one, and they all make sense to me:
> 
> #### Blocking / correctness
> 1. `ResolveMinReplicas` mishandles omitted KEDA `minReplicaCount`
> 
> KEDA defaults an omitted `spec.minReplicaCount` to `0`, but `ResolveMinReplicas` only returns the KEDA floor when `scaledObj.Spec.MinReplicaCount != nil`. If the field is omitted, the code falls through to standalone HPA / deployment replicas and can treat a scale-to-zero KEDA workload as having a higher minimum than it actually does.
> 
> This matters because the PDB floor and EA baseline should reflect KEDA's effective min replica count. Please treat a matching ScaledObject with nil `minReplicaCount` as found with min `0`, rather than falling through.
> 
> 2. Original-min annotation can be overwritten if the surge value changes while a surge is active
> 
> `KEDASurgeApplier.ApplySurge` only skips the ScaledObject update when the existing `evictionSurgeReplicas` annotation already equals the new surge value. If an active surge is first applied from `1 -> 2`, then a later reconcile applies surge `3`, the method recalculates `originalMin` from the already-surged `spec.minReplicaCount=2` and overwrites `OriginalMinReplicasAnnotationKey`. `RevertSurge` would then restore `2` instead of the true original `1`.
> 
> Please preserve the existing original-min annotation once it has been set, and only initialize it when absent. The same pattern appears to exist in the HPA applier as well, so it is worth fixing both.
> 
> #### Medium
> 3. Unsupported KEDA + standalone HPA is returned as a retrying reconcile error
> 
> `detectSurgeApplier` now rejects a ScaledObject and standalone HPA targeting the same deployment, which I agree with. However, the caller returns that error directly from reconcile. Since this is a deterministic unsupported configuration, the controller will keep retrying and users only see the problem in logs.
> 
> Please surface this on `EvictionAutoScaler.Status` instead, for example `Degraded=True` with reason `UnsupportedAutoscalerConfiguration`, and return without continuous requeue churn. I would avoid passing a normal `NoOpSurgeApplier` through the existing scale-up path unless that path is changed not to record a successful scale action.
> 
> #### Low priority cleanup
> 4. Appliers update cached fields before the API write succeeds
> 
> `KEDASurgeApplier` assigns `k.scaledObject = obj` before `client.Update` succeeds. Today this is mostly harmless because appliers are rebuilt per reconcile and the method returns on error, but it makes the in-memory state look committed when the API write failed. Please move the assignment after the successful update or remove it entirely. The same applies to revert and to the HPA applier.
> 
> 5. `CompositeSurgeApplier` and `NoOpSurgeApplier` are now dead code
> 
> `detectSurgeApplier` only returns KEDA, HPA, or deployment appliers. The composite path was useful when the implementation attempted to combine multiple writers, but this PR now makes strategies mutually exclusive and rejects KEDA + standalone HPA. Please remove the unused appliers and the corresponding composite tests rather than leaving dead code plus a future-cleanup comment.
> 
> #### Testing gaps
> Please add focused unit coverage for the edge cases above:
> 
> * KEDA `ScaledObject` with omitted `minReplicaCount` resolves as autoscaler found with min `0`.
> * Reapplying KEDA surge with a different surge value preserves the original-min annotation and reverts to the true original floor. (And the equivalent HPA changed-surge-value case, if the HPA applier has the same overwrite behavior.)
> * ScaledObject + standalone HPA on the same target produces the intended non-retry / degraded status behavior.

addressed all of the concerns

go.mod
```diff
@@ -20,68 +21,80 @@ require (
...
+	go.uber.org/zap v1.27.1
+)
+
+replace (
+	k8s.io/api => k8s.io/api v0.34.3
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.34.3
+	k8s.io/apimachinery => k8s.io/apimachinery v0.34.3
+	k8s.io/client-go => k8s.io/client-go v0.34.3
+	k8s.io/component-base => k8s.io/component-base v0.34.3
+	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.22.4
```
tallaxes: Can we avoid the `replace` block by using an older KEDA minor, e.g. `github.com/kedacore/keda/v2 v2.18.3`?

The current replaces appear to be needed only because KEDA `v2.19.0` has its own Kubernetes/controller-runtime replacement setup. Since dependency `replace` directives are ignored by consumers, this repo has to copy KEDA's replace block into its own `go.mod`. That works, but it is brittle and globally overrides core Kubernetes libraries for the whole controller.

I tested the current typed KEDA code with KEDA `v2.18.3`, Kubernetes/client-go `v0.34.3`, controller-runtime `v0.22.4`, and no replace block; the non-E2E test suite passed. We don't have a specific KEDA `v2.19.0` API requirement (I don't think anything relevant changed in ScaledObject), so I’d prefer pinning to `v2.18.3` and removing the replaces.

(Separately, it would be good to test this with different KEDA versions at runtime ...)

tallaxes (commented):
LGTM, thank you for addressing all feedback! One more thing re component versioning (see below) and I think it is ready to go

go.mod
skuchipudi295: done

tallaxes (approved):

Issue #106, opened 2026-05-04
Controller should not touch StatefulSets


ioanc:
Considering the complexity of the applications running as StatefulSets, the controller should not touch pods related to StatefulSets

PR #109, opened 2026-05-06
skip statefulsets

skuchipudi295:
This pull request updates the `EvictionAutoScaler` controller to explicitly skip handling `StatefulSet` targets, as their ordered pod management is incompatible with the eviction surge strategy. The main change ensures that only supported target types (e.g., Deployments) are processed.

Changes to target handling:

* Updated the `Reconcile` method in `evictionautoscaler_controller.go` to detect when the `EvictionAutoScaler` resource targets a `StatefulSet`, log an informational message, and exit early without attempting to scale or evict pods for that target. (`[internal/controller/evictionautoscaler_controller.goL98-R106](diffhunk://#diff-6c377cb9f9f1e6722559d6e45aa09d2475f9ae85c1b3db39debff112383b69ddL98-R106)`)


skuchipudi295:
fixes #106

PR #115, opened 2026-05-18
modify surge calculation

skuchipudi295:
**EvictionAutoScaler Surge Logic Improvements:**

* Refined the surge scaling logic in the controller to scale up exactly by the number of pods displaced on cordoned nodes, capped by `maxSurge`, and to skip surge when no pods are displaced, preventing unnecessary scaling. [[1]](diffhunk://#diff-6c377cb9f9f1e6722559d6e45aa09d2475f9ae85c1b3db39debff112383b69ddL178-R205) [[2]](diffhunk://#diff-6c377cb9f9f1e6722559d6e45aa09d2475f9ae85c1b3db39debff112383b69ddL191-R214) [[3]](diffhunk://#diff-6c377cb9f9f1e6722559d6e45aa09d2475f9ae85c1b3db39debff112383b69ddL202-R233)
* Added a new helper function `countPodsOnCordoned` to accurately count pods on cordoned nodes, supporting the improved surge logic.

**Testing and Documentation Updates:**

* Updated controller tests to reflect the new surge logic: deployments are not scaled up if no pods are actually displaced on cordoned nodes. [[1]](diffhunk://#diff-ead56681a4d4b68d233bd6c492875dc4f3c5c37c981ac184c1068487b090dd67L231-R235) [[2]](diffhunk://#diff-ead56681a4d4b68d233bd6c492875dc4f3c5c37c981ac184c1068487b090dd67L801-R812) [[3]](diffhunk://#diff-ead56681a4d4b68d233bd6c492875dc4f3c5c37c981ac184c1068487b090dd67R823-R827) [[4]](diffhunk://#diff-ead56681a4d4b68d233bd6c492875dc4f3c5c37c981ac184c1068487b090dd67L888-R897)
* Expanded the `README.md` with a new Networking section, clarifying that the extension does not require ARM endpoints and documenting outbound network requirements for private/restricted clusters. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R13) [[2]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R487-R513)

### DIFF
--- a/internal/controller/evictionautoscaler_controller.go
+++ b/internal/controller/evictionautoscaler_controller.go
@@ -189,39 +189,46 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
 		// surgeTarget = minReplicas + min(maxSurge, displaced)
 		// If displaced == 0 (pods not yet visible on cordoned nodes or race condition),
 		// fall back to full maxSurge so we don't block evictions.
+		// If maxSurge is 0 (not configured), surge by exactly displaced with no cap.
 		maxSurgeTarget := calculateSurge(ctx, target, EvictionAutoScaler.Status.MinReplicas)
 		surgeTarget := EvictionAutoScaler.Status.MinReplicas + displaced
-		if displaced == 0 || surgeTarget > maxSurgeTarget {
-			surgeTarget = maxSurgeTarget
-		}
-
-		if target.GetReplicas() < surgeTarget {
-			logger.Info("No disruptions allowed, scaling up", "pdb", pdb.Name, "lastEviction", EvictionAutoScaler.Spec.LastEviction, "strategy", surgeApplier.Name(), "displaced", displaced, "surgeTarget", surgeTarget)
-
-			// Track blocked eviction if the PDB is blocking the eviction
-			metrics.BlockedEvictionCounter.WithLabelValues(EvictionAutoScaler.Namespace, pdb.Name).Inc()
-
-			// Track scaling opportunity with signal label
-			signalLabel := metrics.GetScalingSignal(pdb)
-			metrics.ScalingOpportunityCounter.WithLabelValues(EvictionAutoScaler.Namespace, EvictionAutoScaler.Spec.TargetName, metrics.ScaleUpAction, signalLabel).Inc()
-
-			err = surgeApplier.ApplySurge(ctx, surgeTarget)
-			if err != nil {
-				logger.Error(err, "failed to apply surge", "kind", EvictionAutoScaler.Spec.TargetKind, "targetname", EvictionAutoScaler.Spec.TargetName, "strategy", surgeApplier.Name())
-				return ctrl.Result{}, err
+		// Cap surgeTarget at maxSurge when displaced exceeds it.
+		// If customer set MaxSurge=0 explicitly, surgeTarget==minReplicas → no surge (they opted out).
+		if displaced == 0 {
+			logger.Info("No pods on cordoned nodes, skipping surge", "pdb", pdb.Name)
+		} else {
+			if surgeTarget > maxSurgeTarget {
+				surgeTarget = maxSurgeTarget
 			}
 
-			// Track actual scaling action
-			metrics.ActualScalingCounter.WithLabelValues(EvictionAutoScaler.Namespace, EvictionAutoScaler.Spec.TargetName, metrics.ScaleUpAction).Inc()
+			if target.GetReplicas() < surgeTarget {
+				logger.Info("No disruptions allowed, scaling up", "pdb", pdb.Name, "lastEviction", EvictionAutoScaler.Spec.LastEviction, "strategy", surgeApplier.Name(), "displaced", displaced, "surgeTarget", surgeTarget)
 
-			// Log the scaling action
-			logger.Info(fmt.Sprintf("Scaled up %s %s/%s to %d replicas (via %s)", EvictionAutoScaler.Spec.TargetKind, target.Obj().GetNamespace(), target.Obj().GetName(), surgeTarget, surgeApplier.Name()))
-			logger.Info(fmt.Sprintf("TargetGeneration moving from %d->%d", EvictionAutoScaler.Status.TargetGeneration, target.Obj().GetGeneration()))
-			// Save ResourceVersion to EvictionAutoScaler status this will cause another reconcile.
-			EvictionAutoScaler.Status.TargetGeneration = target.Obj().GetGeneration()
-			//Do not update EvictionAutoScaler.Status.LastEviction because we need to keep reconciling till scale down
-			ready(&EvictionAutoScaler.Status.Conditions, "Reconciled", "eviction with scale up")
-			return ctrl.Result{RequeueAfter: cooldown}, r.Status().Update(ctx, EvictionAutoScaler)
+				// Track blocked eviction if the PDB is blocking the eviction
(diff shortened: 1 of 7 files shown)


internal/controller/pdb_helpers.go
```diff
@@ -199,3 +199,48 @@ func triggerOnPDBAnnotationChange(e event.UpdateEvent, logger logr.Logger) bool
...
+
+	// Deduplicate node names to minimise Get calls.
+	nodeNames := make(map[string]struct{})
+	for _, pod := range podList.Items {
+		if pod.Spec.NodeName != "" {
+			nodeNames[pod.Spec.NodeName] = struct{}{}
+		}
+	}
+
+	// Fetch each unique node and record whether it is cordoned.
```
paulgmiller: So today we go off of is the node cordoned but there is a small chance in the future we might go off was the pod evicted (like we did with the webhook and may again if that interceptor stuff ever pans out) So might be good to leave a comment that this might change and is just the signal we have now.

internal/controller/pdb_helpers.go
```diff
@@ -199,3 +199,48 @@ func triggerOnPDBAnnotationChange(e event.UpdateEvent, logger logr.Logger) bool
...
+	cordoned := make(map[string]bool, len(nodeNames))
+	for nodeName := range nodeNames {
+		node := &corev1.Node{}
+		if err := c.Get(ctx, k8s_types.NamespacedName{Name: nodeName}, node); err != nil {
```
paulgmiller: nit instead of doign a couple of Get's you might be able to do a list om conrdoned nodes and then just check if a pods node is in that list. Forget if unscheduble works like a label you can select on.

internal/controller/pdb_helpers.go
```diff
@@ -199,3 +199,48 @@ func triggerOnPDBAnnotationChange(e event.UpdateEvent, logger logr.Logger) bool
...
+			if client.IgnoreNotFound(err) != nil {
+				return 0, fmt.Errorf("failed to get node %s: %w", nodeName, err)
+			}
+			// Node not found: pod is being terminated already, don't count it.
```
paulgmiller: worth logging?

internal/controller/pdb_helpers.go
```diff
@@ -199,3 +199,48 @@ func triggerOnPDBAnnotationChange(e event.UpdateEvent, logger logr.Logger) bool
...
+		return 0, fmt.Errorf("invalid PDB selector: %w", err)
+	}
+
+	var podList corev1.PodList
+	if err := c.List(ctx, &podList, client.InNamespace(pdb.Namespace), client.MatchingLabelsSelector{Selector: selector}); err != nil {
+		return 0, fmt.Errorf("failed to list pods for PDB %s: %w", pdb.Name, err)
+	}
+
+	// Deduplicate node names to minimise Get calls.
+	nodeNames := make(map[string]struct{})
```
skuchipudi295: make it concise

internal/controller/evictionautoscaler_controller.go
```diff
@@ -175,37 +175,61 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		// surgeTarget = minReplicas + min(maxSurge, displaced)
+		// If displaced == 0 (pods not yet visible on cordoned nodes or race condition),
+		// fall back to full maxSurge so we don't block evictions.
+		// If maxSurge is 0 (not configured), surge by exactly displaced with no cap.
+		maxSurgeTarget := calculateSurge(ctx, target, EvictionAutoScaler.Status.MinReplicas)
+		surgeTarget := EvictionAutoScaler.Status.MinReplicas + displaced
+		// Cap surgeTarget at maxSurge when displaced exceeds it.
+		// If customer set MaxSurge=0 explicitly, surgeTarget==minReplicas → no surge (they opted out).
+		if displaced == 0 {
+			logger.Info("No pods on cordoned nodes, skipping surge", "pdb", pdb.Name)
```
paulgmiller: good place for an early return 

return ctrl.Result{RequeueAfter: cooldown}, nil

internal/controller/evictionautoscaler_controller.go
```diff
@@ -175,37 +175,61 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		} else {
+			if surgeTarget > maxSurgeTarget {
+				surgeTarget = maxSurgeTarget
```
paulgmiller: log that we've maxed out our surge capacity?

internal/controller/evictionautoscaler_controller.go
paulgmiller: Ah actualyl no this can happen from q requeed event after cooldown if a pod succesfully evicts and we may actually want to scale down.

internal/controller/evictionautoscaler_controller_test.go
```diff
@@ -228,11 +228,11 @@ var _ = Describe("EvictionAutoScaler Controller", func() {
 			//we don't update status of last eviction till
 			Expect(EvictionAutoScaler.Spec.LastEviction.EvictionTime).ToNot(Equal(EvictionAutoScaler.Status.LastEviction.EvictionTime))
 
-			// Verify Deployment scaling if necessary
+			// No real pods exist on a cordoned node, so displaced=0 and no surge fires.
```
paulgmiller, line 233: other thigns we could test. Do we surge up incrmentally as more an more nodes get cordoned?

And do we scale down when some nodes get uncordoned.

internal/controller/evictionautoscaler_controller.go
paulgmiller: Could move displaced = countPodsOnCordoned  out of if condition and then do  pdb.Status.DisruptionsAllowed == 0 && target.GetReplicas() < surgeTarget which is more similiar to the original logic and less chance you'll miss the scaledown cooldown. 

But you also need to answer the question. Do I want to scale down

paulgmiller:
" skip surge when no pods are displaced, preventing unnecessary scaling. " is a bit strange as we should not get this event when no pods are displaced EXCEPT for cooldown requeues in wich case it might make sense to say we're just waiting for cooldown not skipping. 

I would describe this new algorthim as scale up by the maximum of simultaneous pods disrupted during a cooldown and do not scale down till no disruptions for lenght of a cooldown (even if some of those pods already found healthy homes)

paulgmiller (approved):
See my comments about the fact that we're not scaling down  withing the cooldown. I think thats fine but we should get it in readme to not suprise customers.  Some exampels there might be good as well.

internal/controller/evictionautoscaler_controller_test.go
skuchipudi295: we only scale down after cooldown entirely, can add test for that

skuchipudi295:
> See my comments about the fact that we're not scaling down withing the cooldown. I think thats fine but we should get it in readme to not suprise customers. Some exampels there might be good as well.

yup understood, will update readme

paulgmiller:
Not clear to me you need the displaced  == 0 check . if it is zero you'll just skip this 

    if target.GetReplicas() < surgeTarget {


because surgeTarget will be min and you can move on to check if cooldown has expired. 

If we do need to check displaced==0 maybe there should be a test for it? 

https://github.com/Azure/eviction-autoscaler/pull/117

PR #124, opened 2026-05-22
Implement surge revert behavior based on PodDisruptionBudget status

skuchipudi295:
This pull request improves the surge revert logic in the `EvictionAutoScaler` controller to ensure that deployments are only reverted from a surge state when the PodDisruptionBudget (PDB) allows it. Additionally, new tests are added to verify this behavior. The main focus is on preventing premature scale-downs that could violate PDB constraints, thereby improving reliability during eviction scenarios.

**Surge Revert Logic Improvements:**

* The controller now checks `pdb.Status.DisruptionsAllowed` before reverting a deployment from a surge state. If disruptions are not allowed (`DisruptionsAllowed == 0`), it logs a message, updates the status, and requeues the reconciliation after the cooldown period instead of reverting immediately. This ensures that scale-down only happens when it is safe according to the PDB.

**Testing Enhancements:**

* Added a new test suite to `evictionautoscaler_controller_test.go` to verify the surge revert behavior:
  - Tests that the controller does not revert the deployment when `DisruptionsAllowed` is 0 and requeues instead.
  - Tests that the controller reverts the deployment immediately when `DisruptionsAllowed` is greater than 0.

### DIFF
--- a/internal/controller/evictionautoscaler_controller.go
+++ b/internal/controller/evictionautoscaler_controller.go
@@ -245,7 +245,17 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
 		// Track scaling opportunity
 		metrics.ScalingOpportunityCounter.WithLabelValues(EvictionAutoScaler.Namespace, EvictionAutoScaler.Spec.TargetName, metrics.ScaleDownAction, metrics.CooldownElapsedSignal).Inc()
 
-		//okay we aren't at allowed disruptions Revert Target to the original state
+		// Wait until the PDB allows disruptions before reverting.
+		// DisruptionsAllowed > 0 means enough pods are healthy to absorb the scale-down.
+		if pdb.Status.DisruptionsAllowed == 0 {
+			logger.Info("Waiting for PDB to allow disruptions before reverting surge",
+				"pdb", pdb.Name,
+				"target", EvictionAutoScaler.Spec.TargetName)
+			ready(&EvictionAutoScaler.Status.Conditions, "Reconciled", "waiting for PDB to allow disruptions before reverting")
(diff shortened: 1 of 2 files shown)


internal/controller/evictionautoscaler_controller.go
```diff
@@ -245,7 +245,17 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
 		// Track scaling opportunity
 		metrics.ScalingOpportunityCounter.WithLabelValues(EvictionAutoScaler.Namespace, EvictionAutoScaler.Spec.TargetName, metrics.ScaleDownAction, metrics.CooldownElapsedSignal).Inc()
 
-		//okay we aren't at allowed disruptions Revert Target to the original state
+		// Wait until the PDB allows disruptions before reverting.
+		// DisruptionsAllowed > 0 means enough pods are healthy to absorb the scale-down.
+		if pdb.Status.DisruptionsAllowed == 0 {
```
paulgmiller, line 250: Its a little wierd that we check this above on 182 and then check it here again. Might mean its time to break up this function into behavior when when surge is needed an when its not. 

or we could just return early by inverting the logic on  line 201.  f target.GetReplicas() < surgeTarget? to be if replicas >= target { log("waiting for disruption to pass"); return }

I don't think this is wrong its just that this function is getting hard to read so approving with that comment

paulgmiller (approved):
double checking same condition is strange but not wrong.

internal/controller/evictionautoscaler_controller.go
skuchipudi295: I see I actually have a separate reconcile pr.. I just didnt touch in few weeks, may be I can address this comment in that

PR #130, opened 2026-06-08
try and simplify and error on maxsurge=0

paulgmiller:


internal/controller/evictionautoscaler_controller.go
```diff
@@ -186,76 +200,66 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
 			return ctrl.Result{}, countErr
 		}
 
-		// surgeTarget = minReplicas + displaced, capped at minReplicas + maxSurge.
-		// If displaced == 0 the formula yields minReplicas, so no scale-up fires and
-		// we fall through to the cooldown/scale-down path — which is correct.
-		// If maxSurge is 0 (not configured), surgeTarget == minReplicas → no surge (opted out).
```
paulgmiller, line 192: changed this to an error/degrade.

internal/controller/evictionautoscaler_controller.go
```diff
@@ -175,6 +177,18 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 	metrics.EvictionCounter.WithLabelValues(EvictionAutoScaler.Namespace).Inc()
 
+	// surgeTarget = minReplicas + displaced, capped at minReplicas + maxSurge.
+	// If displaced == 0 the formula yields minReplicas, so no scale-up fires and
+	// we fall through to the cooldown/scale-down path — which is correct.
+	maxSurgeTarget := calculateSurge(ctx, target, EvictionAutoScaler.Status.MinReplicas)
+	if maxSurgeTarget == EvictionAutoScaler.Status.MinReplicas {
+		// this should never happen since most eviction autoscalers are generated but possible if manually created
+		// so just error.
+		err := errors.New("max surge 0 eviction autoscaler makes no sense")
```
paulgmiller: could also mark as degraded?

internal/controller/evictionautoscaler_controller.go
paulgmiller, line 193: this always now returns nothing falls through. That lets you simplify futher in the future like so https://github.com/Azure/eviction-autoscaler/pull/131

internal/controller/evictionautoscaler_controller.go
```diff
@@ -175,6 +175,17 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
 		"evictionTime", EvictionAutoScaler.Spec.LastEviction.EvictionTime)
 	metrics.EvictionCounter.WithLabelValues(EvictionAutoScaler.Namespace).Inc()
 
+	// surgeTarget = minReplicas + displaced, capped at minReplicas + maxSurge.
+	// If displaced == 0 the formula yields minReplicas, so no scale-up fires and
+	// we fall through to the cooldown/scale-down path — which is correct.
+	maxSurgeTarget := calculateSurge(ctx, target, EvictionAutoScaler.Status.MinReplicas)
```
skuchipudi295, line 181: Node cordoned → [displaced=1→ we surge (replicas 3→4)
Pod is evicted from the cordoned node — it's now pending/rescheduling elsewhere
Next reconcile: [displaced=0] (pod is no longer on the cordoned node), but [DisruptionsAllowed=0] still (the replacement pod hasn't come up as Ready yet)

internal/controller/evictionautoscaler_controller.go
skuchipudi295: please see above comment

internal/controller/evictionautoscaler_controller.go
skuchipudi295: code returns "waiting for PDB to allow disruptions before reverting" and requeues —  the message is confusing

internal/controller/evictionautoscaler_controller.go
```diff
@@ -186,46 +197,46 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
...
-		if target.GetReplicas() < surgeTarget {
-			logger.Info("No disruptions allowed, scaling up", "pdb", pdb.Name, "lastEviction", EvictionAutoScaler.Spec.LastEviction, "strategy", surgeApplier.Name(), "displaced", displaced, "surgeTarget", surgeTarget)
-
-			// Track blocked eviction if the PDB is blocking the eviction
-			metrics.BlockedEvictionCounter.WithLabelValues(EvictionAutoScaler.Namespace, pdb.Name).Inc()
+		if target.GetReplicas() >= surgeTarget {
+			logger.Info("Waiting for PDB to allow disruptions before reverting surge",
+				"pdb", pdb.Name,
+				"target", EvictionAutoScaler.Spec.TargetName)
+			ready(&EvictionAutoScaler.Status.Conditions, "Reconciled", "waiting for PDB to allow disruptions before reverting")
```
skuchipudi295, line 210: we are inside `if pdb.Status.DisruptionsAllowed == 0{...}`

skuchipudi295: the message looks wrong

paulgmiller: waiting for pb to allow disruptions == "allowed disruptions > 0" 

Maybe 

"Have already scaled up to handle evictions, waiting for PDB to allow disruptions before reverting"

is clearer.  Also added comment //we've scaled up but pdb is still blockign may just be waiting for new pods to become ready

internal/controller/evictionautoscaler_controller.go
paulgmiller: Next reconcile: [displaced=0] (pod is no longer on the cordoned node), but [DisruptionsAllowed=0] still (the replacement pod hasn't come up as Ready yet)

The pod wont leave the cordoned node until the new replacement pods are up and ready because the pdb is protecting it.  If we do take a while to get ready we'll hit the 
```
if target.GetReplicas() >= surgeTarget {
```
That I talk about here.
https://github.com/Azure/eviction-autoscaler/pull/130#discussion_r3385476074

internal/controller/evictionautoscaler_controller.go
paulgmiller: responded to other comments still think is good to not have somethings fall throught this if.

PR #142, opened 2026-07-14
Hardcode eviction-autoscaler for AKS-owned namespaces and reject them in ACTIONED_NAMESPACES

danekeahi:
#### Impact
Makes eviction-autoscaler (EA) always manage AKS-owned namespaces and prevents
customers from configuring EA against them.

Previously `actionedNamespaces` defaulted to `[kube-system]` — a customer-facing Helm
value the customer could remove or repoint at protected namespaces. Since EA is an
AKS-owned extension, whether it manages AKS-owned namespaces (kube-system, flux-system,
gatekeeper-system, etc.) should not be customer-configurable.

#### Changes
- `internal/namespacefilter/nsfilter.go` — add `AKSOwnedNamespaces` (mirrors
  `ProtectedNamespaces` in aks-rp `toolkit/constvalues/automatic/subjects.go`) and
  `IsAKSOwnedNamespace()`. In `Filter()`, AKS-owned namespaces are always enabled,
  independent of `actionedNamespaces` config and the enable annotation.
- `cmd/main.go` — if `ACTIONED_NAMESPACES` contains an AKS-owned namespace, log an
  install error and exit(1) instead of starting the controller.
- `helm/eviction-autoscaler/values.yaml` — `actionedNamespaces: [kube-system]` -> `[]`
  (kube-system is now covered by the hardcoded list, not the customer knob).
- `internal/namespacefilter/nsfilter_test.go` — repoint one existing opt-in test to a
  non-AKS namespace, since AKS-owned namespaces now bypass the annotation path.
- `test/e2e/e2e_test.go` — E2E test compiles and the `actionedNamespaces` scenario was updated so the controller
  no longer fails startup; "kube-system always enabled" is now asserted via the hardcoded
  path rather than via actionedNamespaces.

The `AKSOwnedNamespaces` list is an exact 21/21 match to the current master
`ProtectedNamespaces` (verified programmatically). Additional unit-test coverage is in
a follow-up PR.

#### What tests have been performed
- [x] Unit tests — `go test ./internal/namespacefilter/...` -> `ok`
- [x] Startup rejection — built the manager binary and ran it directly:
  - `ACTIONED_NAMESPACES=kube-system` -> ERROR "may not contain an AKS-owned namespace"
    {namespace: kube-system}, exit 1
  - `ACTIONED_NAMESPACES=production,flux-system` -> ERROR {namespace: flux-system}, exit 1
  - `ACTIONED_NAMESPACES=production` -> allowed ("actionedNamespaces:[production]")
- [x] DS-block resilience (investigation) — simulated Deployment Safeguards rejecting
  the surge write (403 Forbidden) across all three surge strategies (Deployment, HPA,
  KEDA). Each returned the error gracefully and requeues via controller-runtime; no
  panic. Conclusion: EA handles a blocked surge safely (graceful requeue, no crash).

#### Update — test coverage added to this PR

The DS-block investigation and the AKS-owned-namespace behavior are now backed by
committed tests in this PR (previously noted as a follow-up):

- `internal/controller/surge_ds_block_test.go` (new) — unit test that simulates
  Deployment Safeguards rejecting the surge write (403 Forbidden) using a fake-client
  interceptor that forces every `Update` to return `apierrors.NewForbidden(...)`. One
  spec per surge strategy (Deployment, HPA, KEDA); each calls the real `ApplySurge` and
  asserts it (a) returns the error so the reconcile loop requeues and (b) preserves the
  Forbidden type through wrapping (`apierrors.IsForbidden(err)` is true). Ginkgo fails on
  panic, so passing also proves none of the appliers panic. This is the committed version
  of the DS-block resilience investigation described above.
- `test/e2e/e2e_test.go` — three new E2E assertions:
  - **Surge and scale-down in an AKS-owned namespace** — creates a deployment in
    `kube-system` with no enable annotation (AKS-owned namespaces are always managed),
    confirms a PDB and EvictionAutoScaler CR are created automatically, then cordons the
    node to trigger an eviction surge. Asserts the surge happens (the `evictionSurgeReplicas`
    annotation is added and the deployment scales up to 2 available replicas), drains the
    node, and then asserts the scale-down (replicas return to 1 and the surge annotation is
    cleared). The drain only evicts pods labeled `app=nginx-aks-surge`, so it never touches
    real kube-system components.
  - **AKS-owned namespaces ignore the disable annotation** — annotates `kube-system` with
    `eviction-autoscaler.azure.com/enable=false` and verifies via `Consistently` (30s) that
    the PDB still exists, then removes the annotation. Proves an AKS-owned namespace cannot
    be turned off by the customer-facing annotation.
  - **AKS-owned namespace in `actionedNamespaces` is rejected** — installs EA with
    `actionedNamespaces=[kube-system]` and verifies the controller never becomes available,
    a controller pod crashloops, and its logs contain
    `"may not contain an AKS-owned namespace"`. End-to-end proof of the `cmd/main.go`
    startup guard (complements the direct-binary check listed above).

Verification: `go build ./...`, `go vet ./...`, `make lint`, and
`go test ./internal/controller/...` all pass. The E2E additions are compile-verified
locally and exercised in CI.


internal/namespacefilter/nsfilter.go
```diff
@@ -33,6 +64,12 @@ type Reader interface {
 func (n *nsfilter) Filter(ctx context.Context, c Reader, ns string) (bool, error) {
 	logger := ctrl.LoggerFrom(ctx)
 
+	// AKS-owned namespaces are always managed, ignoring config and the enable annotation.
+	if IsAKSOwnedNamespace(ns) {
+		logger.Info("namespace filtering decision", "namespace", ns, "source", "aks-owned", "filtering", true)
+		return true, nil
+	}
```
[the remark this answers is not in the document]
danekeahi: See PR #144

internal/namespacefilter/nsfilter.go
```diff
@@ -14,6 +14,37 @@ import (
 
 const EnableEvictionAutoscalerAnnotationKey = "eviction-autoscaler.azure.com/enable"
 
+// AKSOwnedNamespaces mirrors ProtectedNamespaces in aks-rp
+// (toolkit/constvalues/automatic/subjects.go).
+var AKSOwnedNamespaces = []string{
+	"aks-command",
+	"kube-system",
```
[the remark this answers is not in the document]
danekeahi: made the list unexported (`aksOwnedNamespaces`) so its contents can't be
mutated by other packages; callers use the exported `IsAKSOwnedNamespace()` helper.

internal/namespacefilter/nsfilter.go
```diff
@@ -14,6 +14,38 @@ import (
 
 const EnableEvictionAutoscalerAnnotationKey = "eviction-autoscaler.azure.com/enable"
 
+// aksOwnedNamespaces mirrors ProtectedNamespaces in aks-rp
+// (toolkit/constvalues/automatic/subjects.go). It is intentionally unexported so its
+// contents cannot be mutated by other packages; use IsAKSOwnedNamespace to query it.
+var aksOwnedNamespaces = []string{
```
skuchipudi295, line 24: please add a comment here saying check against rp code to keep it same before every ext release

danekeahi: Added a note above the list reminding maintainers that it's a hardened copy of the RP's ProtectedNamespaces, and to check it against aks-rp subjects.go and re-sync before every extension release.

test/e2e/e2e_test.go
```diff
@@ -1517,5 +1536,208 @@ var _ = Describe("controller", Ordered, func() {
...
+			err = clientset.Update(ctx, &node)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("verifying the deployment gets the evictionSurgeReplicas annotation (surge happened)")
+			EventuallyWithOffset(1, func() error {
+				var dep appsv1.Deployment
+				if err := clientset.Get(ctx, client.ObjectKey{Namespace: aksNs, Name: depName}, &dep); err != nil {
+					return err
+				}
+				if _, ok := dep.Annotations["evictionSurgeReplicas"]; !ok {
```
skuchipudi295: use a constant for this

skuchipudi295: can use same constant used in actual code

internal/controller/surge_ds_block_test.go
```diff
@@ -0,0 +1,93 @@
...
+		},
+	})
+}
+
+// When Deployment Safeguards blocks the surge write, each surge strategy must return the
+// error (so the reconcile loop requeues) and must not panic. A panic here would fail the
+// spec because Ginkgo treats panics as failures.
+var _ = Describe("Surge appliers when Deployment Safeguards blocks the write", func() {
+	var ctx context.Context
+
```
skuchipudi295: this test should be for eviction autoscaler controller.. please move this file into separate pr

skuchipudi295 (changes requested):

test/e2e/e2e_test.go
danekeahi: Now uses controller.EvictionSurgeReplicasAnnotationKey from the controller package. (Left the pre-existing literals in the older tests as-is to keep this PR focused)

internal/controller/surge_ds_block_test.go
danekeahi: Moved to PR #151.

skuchipudi295 (approved):

PR #147, opened 2026-07-18
feat: add surge-override annotation to decouple drain surge from maxSurge

NamanBhat:
## What / why

The eviction autoscaler derives its drain-time surge budget from the target's rolling-update `maxSurge`. Workloads that pin `maxSurge: 0` — a common safe-deployment requirement so app rollouts never create extra pods — get **no** drain-time surge at all, so `calculateSurge` returns an error and the controller cannot add replicas to relieve a PDB-blocked drain. This disables the autoscaler for exactly the workloads that most need disruption headroom.

## Change

Add an annotation `eviction-autoscaler.azure.com/surge-override` on the target workload (Deployment/StatefulSet). When present it **always wins** over `maxSurge`, using the same int-or-percentage semantics. A workload can keep `maxSurge: 0` for rollouts while still allowing a bounded, drain-only surge that the controller reverts when the drain finishes.

- Add `SurgeOverrideAnnotationKey` constant.
- `calculateSurge` reads the override first, falling back to `maxSurge`.
- Extract `surgeFromValue` helper for the int/percentage math.
- Unit tests: int override with `maxSurge=0`, override wins over non-zero `maxSurge`, percentage, no-RollingUpdate, zero (→ `errMaxSurgeZero`), invalid (→ `errInvalidPercentage`).
- Document the annotation in the README under "How Surge Sizing Works".

## Testing

- `go build ./...` clean.
- Full controller suite (`go test ./internal/controller/...`, envtest) passes, including the new override tests.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -196,9 +196,10 @@ func main() {
 	setupLog.Info("PDB creation configuration", "pdbCreate", pdbCreate)
 
 	if err = (&controllers.EvictionAutoScalerReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
-		Filter: nsfilter,
+		Client:   mgr.GetClient(),
+		Scheme:   mgr.GetScheme(),
+		Filter:   nsfilter,
+		Recorder: mgr.GetEventRecorderFor("eviction-autoscaler"),
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "EvictionAutoScaler")
 		os.Exit(1)
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -19,7 +19,6 @@ package main
 import (
 	"crypto/tls"
 	"flag"
-	"fmt"
 	"os"
 	"strconv"
 	"strings"
@@ -210,35 +209,11 @@ func main() {
 	}
 	setupLog.Info("Zero-maxSurge override configuration", "zeroSurgeOverride", zeroSurgeOverride)
 
-	// Parse ENABLE_EVENT_RECORDING environment variable (defaults to false if not set).
-	// Controls whether the controller emits Kubernetes events at all — an
-	// installation-time observability decision, orthogonal to feature flags. When off,
-	// the Recorder is left nil and the `if r.Recorder != nil` guards throughout the
-	// controller skip all event emission (structured logs + metrics remain the primary
-	// signal). Strictly validated to "true"/"false"/empty so a typo fails fast.
-	eventRecordingEnabled := false
-	switch v := os.Getenv("ENABLE_EVENT_RECORDING"); v {
-	case "", "false":
-		eventRecordingEnabled = false
-	case "true":
-		eventRecordingEnabled = true
-	default:
-		setupLog.Error(fmt.Errorf("invalid value %q for ENABLE_EVENT_RECORDING (must be \"true\" or \"false\")", v),
-			"Failed to parse ENABLE_EVENT_RECORDING env variable")
-		os.Exit(1)
-	}
-	setupLog.Info("Event recording configuration", "eventRecordingEnabled", eventRecordingEnabled)
-
 	evictionAutoScalerReconciler := &controllers.EvictionAutoScalerReconciler{
-		Client:                  mgr.GetClient(),
-		Scheme:                  mgr.GetScheme(),
-		Filter:                  nsfilter,
-		ZeroSurgeOverride:       zeroSurgeOverride,
-	}
-	// Only wire the event recorder when event recording is enabled; otherwise the
-	// nil Recorder disables all event emission via the existing guards.
-	if eventRecordingEnabled {
-		evictionAutoScalerReconciler.Recorder = mgr.GetEventRecorderFor("eviction-autoscaler")
+		Client:            mgr.GetClient(),
+		Scheme:            mgr.GetScheme(),
+		Filter:            nsfilter,
+		ZeroSurgeOverride: zeroSurgeOverride,
 	}
 	if err = evictionAutoScalerReconciler.SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "EvictionAutoScaler")
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"crypto/tls"
 	"flag"
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
@@ -195,12 +196,71 @@ func main() {
 	}
 	setupLog.Info("PDB creation configuration", "pdbCreate", pdbCreate)
 
-	if err = (&controllers.EvictionAutoScalerReconciler{
-		Client:   mgr.GetClient(),
-		Scheme:   mgr.GetScheme(),
-		Filter:   nsfilter,
-		Recorder: mgr.GetEventRecorderFor("eviction-autoscaler"),
-	}).SetupWithManager(mgr); err != nil {
+	// Parse SURGE_OVERRIDE_ENABLED environment variable (defaults to false if not set).
+	// Fleet-wide opt-in for honoring the drain-time surge-override annotation; when
+	// off, the annotation is a no-op and surge sizing falls through to maxSurge.
+	// Strictly validated to "true"/"false"/empty so a typo (e.g. "yes") fails fast at
+	// startup rather than silently disabling the feature.
+	surgeOverrideEnabled := false
+	switch v := os.Getenv("SURGE_OVERRIDE_ENABLED"); v {
+	case "", "false":
+		surgeOverrideEnabled = false
+	case "true":
+		surgeOverrideEnabled = true
+	default:
+		setupLog.Error(fmt.Errorf("invalid value %q for SURGE_OVERRIDE_ENABLED (must be \"true\" or \"false\")", v),
+			"Failed to parse SURGE_OVERRIDE_ENABLED env variable")
+		os.Exit(1)
+	}
+	setupLog.Info("Surge override configuration", "surgeOverrideEnabled", surgeOverrideEnabled)
+
+	// Parse SURGE_OVERRIDE_MAX_PERCENT environment variable (defaults to 25 if not set).
+	// Bounds the free-form surge-override annotation as a fleet-wide safety net; must be
+	// a positive integer. Operators can raise it to allow larger drain-time surges.
+	surgeOverrideMaxPercent := int32(25)
+	if v := os.Getenv("SURGE_OVERRIDE_MAX_PERCENT"); v != "" {
+		p, err := strconv.Atoi(v)
+		if err != nil || p <= 0 {
+			setupLog.Error(fmt.Errorf("SURGE_OVERRIDE_MAX_PERCENT must be a positive integer, got %q", v),
+				"Failed to parse SURGE_OVERRIDE_MAX_PERCENT env variable")
+			os.Exit(1)
+		}
+		surgeOverrideMaxPercent = int32(p)
+	}
+	setupLog.Info("Surge override max percent configuration", "surgeOverrideMaxPercent", surgeOverrideMaxPercent)
+
+	// Parse ENABLE_EVENT_RECORDING environment variable (defaults to false if not set).
+	// Controls whether the controller emits Kubernetes events at all — an
+	// installation-time observability decision, orthogonal to feature flags. When off,
+	// the Recorder is left nil and the `if r.Recorder != nil` guards throughout the
+	// controller skip all event emission (structured logs + metrics remain the primary
+	// signal). Strictly validated to "true"/"false"/empty so a typo fails fast.
+	eventRecordingEnabled := false
+	switch v := os.Getenv("ENABLE_EVENT_RECORDING"); v {
+	case "", "false":
+		eventRecordingEnabled = false
+	case "true":
+		eventRecordingEnabled = true
+	default:
+		setupLog.Error(fmt.Errorf("invalid value %q for ENABLE_EVENT_RECORDING (must be \"true\" or \"false\")", v),
+			"Failed to parse ENABLE_EVENT_RECORDING env variable")
+		os.Exit(1)
+	}
+	setupLog.Info("Event recording configuration", "eventRecordingEnabled", eventRecordingEnabled)
+
+	evictionAutoScalerReconciler := &controllers.EvictionAutoScalerReconciler{
+		Client:                  mgr.GetClient(),
+		Scheme:                  mgr.GetScheme(),
+		Filter:                  nsfilter,
+		SurgeOverrideEnabled:    surgeOverrideEnabled,
+		SurgeOverrideMaxPercent: surgeOverrideMaxPercent,
+	}
+	// Only wire the event recorder when event recording is enabled; otherwise the
+	// nil Recorder disables all event emission via the existing guards.
+	if eventRecordingEnabled {
+		evictionAutoScalerReconciler.Recorder = mgr.GetEventRecorderFor("eviction-autoscaler")
+	}
+	if err = evictionAutoScalerReconciler.SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "EvictionAutoScaler")
 		os.Exit(1)
 	}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -196,41 +196,19 @@ func main() {
 	}
 	setupLog.Info("PDB creation configuration", "pdbCreate", pdbCreate)
 
-	// Parse OVERRIDE_ZERO_MAXSURGE environment variable (defaults to false if not set).
-	// Fleet-wide opt-in: when on, a workload whose maxSurge resolves to 0 (an explicit
-	// maxSurge: 0 or a Recreate strategy) is surged by SURGE_OVERRIDE_MAX_PERCENT of
-	// minReplicas during a drain instead of being unable to surge. When off, such a
-	// workload degrades as today. Strictly validated to "true"/"false"/empty so a typo
-	// (e.g. "yes") fails fast at startup rather than silently disabling the feature.
-	overrideZeroMaxSurge := false
-	switch v := os.Getenv("OVERRIDE_ZERO_MAXSURGE"); v {
-	case "", "false":
-		overrideZeroMaxSurge = false
-	case "true":
-		overrideZeroMaxSurge = true
-	default:
-		setupLog.Error(fmt.Errorf("invalid value %q for OVERRIDE_ZERO_MAXSURGE (must be \"true\" or \"false\")", v),
-			"Failed to parse OVERRIDE_ZERO_MAXSURGE env variable")
+	// Parse ZERO_SURGE_OVERRIDE environment variable (unset/empty ⇒ feature off).
+	// Fleet-wide, install-time knob: when set, a workload whose maxSurge resolves to
+	// 0 (an explicit maxSurge: 0, a Recreate strategy, or an unset RollingUpdate) is
+	// surged by this value during a drain instead of degrading. The value is an
+	// int-or-percentage, mirroring Kubernetes maxSurge — "25%" of minReplicas
+	// (rounded up) or an absolute "10". A value that resolves to zero ("0"/"0%")
+	// leaves the feature off; a negative or malformed value fails fast at startup.
+	zeroSurgeOverride, zsErr := controllers.ParseZeroSurgeOverride(os.Getenv("ZERO_SURGE_OVERRIDE"))
+	if zsErr != nil {
+		setupLog.Error(zsErr, "Failed to parse ZERO_SURGE_OVERRIDE env variable")
 		os.Exit(1)
 	}
-	setupLog.Info("Zero-maxSurge override configuration", "overrideZeroMaxSurge", overrideZeroMaxSurge)
-
-	// Parse SURGE_OVERRIDE_MAX_PERCENT environment variable (defaults to 25 if not set).
-	// The fleet-wide drain surge, as a percent of minReplicas, applied when
-	// OVERRIDE_ZERO_MAXSURGE is on and a workload's maxSurge resolves to 0. Must be a
-	// positive integer (e.g. set to 10 at install time). The actual surge stays
-	// demand-driven and is capped at this amount, so larger drains proceed in waves.
-	surgeOverrideMaxPercent := int32(25)
-	if v := os.Getenv("SURGE_OVERRIDE_MAX_PERCENT"); v != "" {
-		p, err := strconv.Atoi(v)
-		if err != nil || p <= 0 {
-			setupLog.Error(fmt.Errorf("SURGE_OVERRIDE_MAX_PERCENT must be a positive integer, got %q", v),
-				"Failed to parse SURGE_OVERRIDE_MAX_PERCENT env variable")
-			os.Exit(1)
-		}
-		surgeOverrideMaxPercent = int32(p)
-	}
-	setupLog.Info("Surge override max percent configuration", "surgeOverrideMaxPercent", surgeOverrideMaxPercent)
+	setupLog.Info("Zero-maxSurge override configuration", "zeroSurgeOverride", zeroSurgeOverride)
 
 	// Parse ENABLE_EVENT_RECORDING environment variable (defaults to false if not set).
 	// Controls whether the controller emits Kubernetes events at all — an
@@ -255,8 +233,7 @@ func main() {
 		Client:                  mgr.GetClient(),
 		Scheme:                  mgr.GetScheme(),
 		Filter:                  nsfilter,
-		OverrideZeroMaxSurge:    overrideZeroMaxSurge,
-		SurgeOverrideMaxPercent: surgeOverrideMaxPercent,
+		ZeroSurgeOverride:       zeroSurgeOverride,
 	}
 	// Only wire the event recorder when event recording is enabled; otherwise the
 	// nil Recorder disables all event emission via the existing guards.
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -196,27 +196,30 @@ func main() {
 	}
 	setupLog.Info("PDB creation configuration", "pdbCreate", pdbCreate)
 
-	// Parse SURGE_OVERRIDE_ENABLED environment variable (defaults to false if not set).
-	// Fleet-wide opt-in for honoring the drain-time surge-override annotation; when
-	// off, the annotation is a no-op and surge sizing falls through to maxSurge.
-	// Strictly validated to "true"/"false"/empty so a typo (e.g. "yes") fails fast at
-	// startup rather than silently disabling the feature.
-	surgeOverrideEnabled := false
-	switch v := os.Getenv("SURGE_OVERRIDE_ENABLED"); v {
+	// Parse OVERRIDE_ZERO_MAXSURGE environment variable (defaults to false if not set).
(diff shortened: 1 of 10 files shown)


internal/controller/evictionautoscaler_controller.go
```diff
@@ -323,13 +336,27 @@ var (
...
 func calculateSurge(_ context.Context, target Surger, minrepicas int32) (int32, error) {
+	// An explicit surge-override annotation on the target always wins over maxSurge,
+	// so a workload can keep maxSurge=0 for its rollout strategy yet still surge on drain.
+	if ann := target.Obj().GetAnnotations(); ann != nil {
+		if override, ok := ann[SurgeOverrideAnnotationKey]; ok {
+			return surgeFromValue(intstr.Parse(override), minrepicas)
+		}
+	}
+	return surgeFromValue(target.GetMaxSurge(), minrepicas)
+}
```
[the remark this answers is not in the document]
NamanBhat: Fixed in 846fcd6 — renamed `minrepicas` -> `minReplicas` throughout `calculateSurge` and `surgeFromValue`.

internal/controller/evictionautoscaler_controller.go
```diff
@@ -323,13 +336,27 @@ var (
 	errInvalidPercentage = errors.New("invalid surge percentage")
 )
 
-// calculateSurge returns the maximum replica count after surge (minReplicas + maxSurge).
+// calculateSurge returns the maximum replica count after surge (minReplicas + surge).
+// The surge amount is taken from the SurgeOverrideAnnotationKey annotation on the
+// target when present (it always wins), otherwise from the target's maxSurge.
 // Returns a sentinel error to distinguish:
-//   - errMaxSurgeZero: maxSurge resolves to 0 (explicitly set or not configured)
+//   - errMaxSurgeZero: the surge amount resolves to 0 (explicitly set or not configured)
```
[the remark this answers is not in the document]
NamanBhat: Updated approach (as of 536f389): after discussion we decided to **keep** the `errMaxSurgeZero` sentinel and its original message unchanged, and leave the existing maxSurge path untouched. Instead, only the new surge-override path now wraps the error with the annotation value as context, e.g. `surge-override annotation "0": maxSurge is 0; eviction autoscaler cannot surge`. That gives operators the source distinction without changing existing behavior; errors still unwrap to the same sentinels via `errors.Is`.

NamanBhat:
Thanks for the review! Addressed the feedback (final state in 536f389):

**Input validation (`surgeFromValue`)** — the low-confidence note about non-`%` strings and negatives was valid, now handled:
- A string surge value without a `%` suffix (e.g. `"10"`) is rejected with `errInvalidPercentage` instead of being silently interpreted as a percentage. Note: a numeric `maxSurge` (e.g. `maxSurge: 2`) is stored by Kubernetes as an intstr **Int** and goes through the Int branch, so it is never affected — the `%` guard only catches a malformed String value.
- Negative int or percentage values are rejected with a new `errNegativeSurge` sentinel, so the computed target can never drop below `minReplicas`.
- Added unit tests for the no-`%` string, negative int, negative percentage, and negative override cases.

**Naming/message** — renamed `minrepicas` → `minReplicas`. For the "maxSurge is 0" message, we kept the `errMaxSurgeZero` sentinel and its original message unchanged and left the existing maxSurge path as-is; the new surge-override path wraps the error with the annotation value so operators can tell the 0/invalid value came from the override (replied inline).

`go build`, `go vet`, and the full controller suite (94/94 specs) pass locally.

NamanBhat:
@microsoft-github-policy-service agree company="Microsoft"

internal/controller/evictionautoscaler_controller.go
```diff
@@ -321,34 +334,69 @@ func (r *EvictionAutoScalerReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+	return surgeFromValue(target.GetMaxSurge(), minReplicas)
+}
 
-	surge := target.GetMaxSurge()
+// surgeFromValue resolves an int-or-percentage surge value against minReplicas.
+// An int is added directly; a percentage (a string ending in "%") is applied to
+// minReplicas and rounded up. A zero value yields errMaxSurgeZero; a negative value
+// yields errNegativeSurge; a string that is not a valid "<n>%" percentage yields
+// errInvalidPercentage.
+func surgeFromValue(surge intstr.IntOrString, minReplicas int32) (int32, error) {
```
skuchipudi295, line 434: The annotation decouples surge from maxSurge.


Replace the [surge-override] annotation with a single controller config toggle, e.g. OVERRIDE_ZERO_MAXSURGE=true (env var, parsed in [main.go], passed as a field on EvictionAutoScalerReconciler — same pattern as PDB_CREATE).
When enabled and surge resolves to 0 (maxSurge: 0/unset), use a hardcoded 25% ceiling instead of returning [errMaxSurgeZero].
Off by default → preserves today's degrade behavior.
Rationale: maxSurge is only the cap; the actual surge is demand-driven by displaced pods ([surgeTarget = minReplicas + displaced], capped at [minReplicas + maxSurgeTarget]. So 25% doesn't need to be per-deployment tunable — it's just a fleet-wide batch-size ceiling. Drains larger than 25% proceed in waves, same as the existing incremental scale-up. Correctness is unchanged; 25% only bounds throughput.

Two asks if we go this route:

Keep the [errMaxSurgeZero] → degrade path as the default (toggle off).
Emit an event/log when the override raises the ceiling, since we're overriding an author's explicit maxSurge: 0 fleet-wide.
The [surgeFromValue] helper and its tests here are still useful and can back the 25% math.

skuchipudi295 (changes requested):

internal/controller/evictionautoscaler_controller.go
NamanBhat: Thanks Sushma — good push. This isn''t really a *"make `maxSurge: 0` surge"* toggle; it''s a **drain-time surge cap override, intentionally decoupled from the rollout `maxSurge`**.

Rollout surge and drain surge are different operational events: a workload can want `maxSurge: 0` for **app rollouts** (serial, no extra pods during a version change) yet still want a **small, bounded surge during node maintenance**. Today those share one knob; this annotation separates them.

On "hardcoded 25%, fleet-wide": you''re right it''s only a **cap** and the surge is demand-driven (`minReplicas + displaced`, waves for bigger drains). That''s exactly why it should be **owned and tuned by the workload/namespace team**, not a fleet constant — concrete case: a team with `maxSurge: 0` on a ~500-replica deployment often wants a **small absolute drain cap, e.g. 10 extra pods**, not 25% (= 125 pods), which is the disruption they set `maxSurge: 0` to avoid. The annotation takes an int **or** a percentage precisely so each workload can tune the batch size to its capacity — same reason `maxSurge` lives on the workload, not the controller.

Where I fully agree:
- **Off/degrade by default** — already the case (absent annotation → unchanged `errMaxSurgeZero`).
- Keep `surgeFromValue` + tests.
- **Event/log when the drain cap overrides `maxSurge`** — done in 586a429: a `DrainSurgeOverride` event on the target workload + a structured log when the override drives a surge.

I''ve also reframed the docs/naming as a *drain-time surge cap override* to make the decoupling explicit.

skuchipudi295 (changes requested):
## Gate behind a controller-level feature flag

The surge-override annotation is currently unconditionally honored — any workload owner who can annotate their Deployment/StatefulSet activates new surge behavior without Cosmic opting in. This needs a controller-level gate.

### Required changes

**1. Add a controller-level env var (e.g. `SURGE_OVERRIDE_ENABLED`, off by default)**

Same pattern as `PDB_CREATE`: env var parsed in `main.go`, passed as a bool field on `EvictionAutoScalerReconciler`. When disabled (default), the annotation is a no-op — `calculateSurge` falls straight through to `maxSurge` as it does today.

```go
// main.go
surgeOverrideEnabled := os.Getenv("SURGE_OVERRIDE_ENABLED") == "true"

// reconciler
reconciler := &controllers.EvictionAutoScalerReconciler{
    Client:               mgr.GetClient(),
    Scheme:               mgr.GetScheme(),
    Filter:               nsfilter,
    SurgeOverrideEnabled: surgeOverrideEnabled,
}
if surgeOverrideEnabled {
    reconciler.Recorder = mgr.GetEventRecorderFor("eviction-autoscaler")
}
```

```go
// calculateSurge — only read the annotation when the flag is on
func calculateSurge(ctx context.Context, target Surger, minReplicas int32, surgeOverrideEnabled bool) (int32, error) {
    if surgeOverrideEnabled {
        if override, ok := surgeOverrideValue(target); ok {
            // ...
        }
    }
    return surgeFromValue(target.GetMaxSurge(), minReplicas)
}
```

Rationale: Cosmic controls whether tenants can activate drain-time surge. Workload annotations alone should not be able to change controller behavior without an explicit fleet-level opt-in.

**2. Gate the event recorder behind the same flag**

The `DrainSurgeOverride` event emission and the `Recorder` field itself should only be wired when the flag is enabled. At fleet scale (hundreds of annotated workloads draining simultaneously), event writes compete with the controller's primary scale-up PATCHes for API server bandwidth, and the client-go recorder's internal rate limiter (~25 burst) means events silently drop under load anyway. Gating ensures we only pay this cost when the feature is deliberately enabled.

When the flag is off:
- No `Recorder` on the reconciler (nil, zero cost)
- No event emission path is reachable
- Structured log line is also skipped (no override to log about)

When the flag is on:
- `Recorder` is set, events fire when an override-driven surge actually executes
- Structured log + event together give both controller-side and workload-side observability

**3. Keep `surgeFromValue` helper + tests as-is**

The refactored helper, negative-surge validation, and bare-number rejection are good improvements regardless of the feature gate. Those can stay.

---

TL;DR: Feature needs to be opt-in at the controller level (Cosmic decides), not opt-in at the workload level (tenants decide). Default off, same pattern as `PDB_CREATE`.

skuchipudi295:
One more ask on top of the review:

**Testing coverage needed:**

- **E2E test**: Add an envtest-based integration test (or extend the existing controller suite) that exercises the full reconcile path with the feature flag enabled + annotation set. Specifically:
  - Workload with `maxSurge: 0` + surge-override annotation → controller surges correctly
  - Same scenario with feature flag *disabled* → annotation is ignored, controller degrades as today
  - Drain completes → surge is reverted back to original replicas

- **More unit tests for `surgeFromValue`**: The existing override tests are good, but add coverage for:
  - Boundary: override = `"1"` with `minReplicas = 1` (minimal surge)
  - Large percentage: `"100%"` (doubles replicas)
  - Percentage rounding: values that exercise `math.Ceil` (e.g. `"33%"` with `minReplicas = 10` → ceil(3.3) = 4)
  - Feature flag off + annotation present → falls through to maxSurge (this is the critical behavioral test for the gate)

The feature gate is the main behavioral change here — we need tests proving the gate actually works (annotation ignored when flag is off).

skuchipudi295:
Also: validate the env var in `main.go` at startup. If `SURGE_OVERRIDE_ENABLED` is set to something other than `"true"` or `"false"` (or empty/unset), fail fast with a clear error rather than silently defaulting to off. Same defensive pattern you'd want for any feature flag — avoids a typo like `SURGE_OVERRIDE_ENABLED=yes` silently disabling the feature when the operator thinks it's on.

```go
surgeOverrideRaw := os.Getenv("SURGE_OVERRIDE_ENABLED")
var surgeOverrideEnabled bool
switch surgeOverrideRaw {
case "", "false":
    surgeOverrideEnabled = false
case "true":
    surgeOverrideEnabled = true
default:
    setupLog.Error(fmt.Errorf("invalid value %q for SURGE_OVERRIDE_ENABLED (must be \"true\" or \"false\")"), "")
    os.Exit(1)
}
setupLog.Info("Surge override configuration", "surgeOverrideEnabled", surgeOverrideEnabled)
```

skuchipudi295:
One more: the README section ("Drain-Time Surge Cap Override") needs to be updated once the controller flag is added. Currently it reads as if the annotation is unconditionally active — just annotate your workload and you get drain surge.

After the gate is in place, the docs should note:
- The feature requires `SURGE_OVERRIDE_ENABLED=true` on the controller (fleet-wide, off by default)
- The annotation is a no-op unless the controller flag is enabled
- Add the env var to any "Enable/config summary" table (same pattern as #148's README section for `ENABLE_PDB_FLOOR_MUTATION`)

Otherwise operators will annotate their workloads, wonder why nothing happens, and open a support ticket.

skuchipudi295:
Additional requirement: **Event recording should be behind a dedicated, separate controller flag (`ENABLE_EVENT_RECORDING`, default off).**

This is orthogonal to feature flags — it's an installation-time decision about observability verbosity, not about feature behavior. Implementation:

```go
// main.go
eventRecordingEnabled := os.Getenv("ENABLE_EVENT_RECORDING") == "true"

reconciler := &controllers.EvictionAutoScalerReconciler{
    Client: mgr.GetClient(),
    Scheme: mgr.GetScheme(),
    Filter: nsfilter,
}
if eventRecordingEnabled {
    reconciler.Recorder = mgr.GetEventRecorderFor("eviction-autoscaler")
}
```

Then the existing `if r.Recorder != nil` guard throughout the controller naturally gates all event emission — no per-feature event checks needed.

**Why separate from feature flags:**
- Single knob for all event noise across all features (surge override, PDB mutation, future features)
- Operators who know their fleet is large enough that events are noise can disable globally at install time
- Feature flags control *behavior*; this flag controls *observability verbosity* — orthogonal concerns
- Default off: structured logs + metrics are the primary path; events are opt-in for operators who actually use `kubectl describe`

Validate this the same way as other env vars — fail fast on invalid values.

skuchipudi295:
**Critical: cap the surge-override value at 25% of minReplicas.**

Currently `surgeFromValue` accepts any positive int32 or percentage with no upper bound. Since the annotation is free-form text (no admission webhook validating it), a typo like `"100"` on a 500-replica deployment → `surgeTarget = 600` → potential quota exhaustion, scheduler pressure, and exactly the kind of uncontrolled scale event that `maxSurge: 0` was trying to prevent.

Kubernetes defaults `maxSurge` to 25% for a reason — it's a safe batch-size ceiling. The override should respect that same ceiling:

```go
const maxSurgeOverrideCap = 0.25 // 25% of minReplicas, matching K8s default maxSurge

func surgeFromValue(surge intstr.IntOrString, minReplicas int32) (int32, error) {
    // ... existing parsing ...

    // Cap the resolved surge amount at 25% of minReplicas (rounded up).
    cap := int32(math.Ceil(float64(minReplicas) * maxSurgeOverrideCap))
    if cap < 1 {
        cap = 1
    }
    if resolvedSurge > cap {
        resolvedSurge = cap
    }
    return minReplicas + resolvedSurge, nil
}
```

**Why:**
- The annotation bypasses K8s API server validation (unlike the real `maxSurge` field which is validated at admission)
- An operator typing `"100"` when they meant `"10"` shouldn't be able to trigger a 100-pod surge
- 25% matches Kubernetes' own default and keeps drain batches bounded
- If someone truly needs more than 25%, they should be making that case explicitly (and we can revisit the cap)

Also: consider `strings.TrimSpace` on the annotation value before `intstr.Parse` — operators will typo with leading/trailing spaces and get a confusing "not a percentage" error instead of a clear parse failure.

internal/controller/evictionautoscaler_controller.go
NamanBhat: Agreed on the safety concern — the annotation is free-form and bypasses API-server admission validation, so an unbounded value is a real risk. I implemented it as a *configurable* cap rather than a hardcoded 25%: `SURGE_OVERRIDE_MAX_PERCENT` (default `25`) bounds the override-resolved surge amount (`ceil(pct × minReplicas)`, floored at 1), applied **only** to the override path — the admission-validated `maxSurge` path is never capped. Defaulting to 25% gives the safety net you asked for out of the box, while letting an operator raise it fleet-wide preserves the per-workload tunability this annotation exists for (e.g. a large `maxSurge: 0` deployment pinning a small absolute cap). Also added `strings.TrimSpace` on the value before parsing. Done in `92c8bc6`.

NamanBhat:
**Re: gate behind a controller feature flag** — Done in `92c8bc6`. Added `SURGE_OVERRIDE_ENABLED` (env, off by default, fail-fast) as `EvictionAutoScalerReconciler.SurgeOverrideEnabled`; when off, `calculateSurge` ignores the annotation and falls through to `maxSurge` (prior behavior preserved). The override event/log path is also gated on the flag. Per your later comment I gated the event *recorder* behind the dedicated `ENABLE_EVENT_RECORDING` flag instead of this one (see that reply). Kept `surgeFromValue` + its tests.

NamanBhat:
**Re: testing coverage** — Done. Controller-suite (envtest) coverage: flag on → surges + emits `DrainSurgeOverride`; flag off → annotation ignored + Degraded (`UnsupportedAutoscalerConfiguration`), no scale-up; drain completes → reverts to original replicas; surge with a nil recorder (no panic, no event). Unit tests: boundary `"1"` with minReplicas 1, `"100%"`, `"33%"` ceil rounding (→4 on minReplicas 10), and flag-off fall-through. All green locally.

NamanBhat:
**Re: startup env validation** — Done. `SURGE_OVERRIDE_ENABLED` and `ENABLE_EVENT_RECORDING` use a strict `""`/`"false"`/`"true"` switch with fail-fast (`os.Exit(1)`) on anything else; `SURGE_OVERRIDE_MAX_PERCENT` requires a positive integer. Resolved values are logged at startup. I scoped the strict validation to the new flags; the existing `PDB_CREATE`/`ENABLED_BY_DEFAULT` continue to use `strconv.ParseBool` as before.

NamanBhat:
**Re: README** — Done. The env-var list now documents `SURGE_OVERRIDE_ENABLED`, `SURGE_OVERRIDE_MAX_PERCENT`, and `ENABLE_EVENT_RECORDING`. The "Drain-Time Surge Cap Override" section states the feature is off by default and the annotation is a no-op unless the controller flag is enabled, and the event line notes it emits only when `ENABLE_EVENT_RECORDING=true`.

NamanBhat:
**Re: dedicated event-recording flag** — Done. Added `ENABLE_EVENT_RECORDING` (off by default, strict validation). The `Recorder` is wired on the reconciler only when the flag is enabled, so the existing `if r.Recorder != nil` guards gate all event emission fleet-wide — no per-feature event checks. This supersedes the recorder-gating from the first comment.

skuchipudi295:
@NamanBhat  QQ, do we have a validation check that only use if configurable knobs are set appropriately and maxSruge on deployments = 0?

NamanBhat:
Good question. Two parts:

**Knob validation ? yes.** The override is only honored when `SURGE_OVERRIDE_ENABLED=true` (fleet-wide opt-in, strictly validated at startup ? fail-fast on anything but `true`/`false`). The resolved override surge is then bounded by `SURGE_OVERRIDE_MAX_PERCENT` (default 25% of `minReplicas`), and `surgeFromValue` rejects negative, zero, and bare-number values. So a misconfigured knob or an invalid annotation value can't produce an unbounded surge.

**`maxSurge == 0` as a precondition ? intentionally not.** The annotation's whole purpose is to **decouple drain-time surge from the rollout `maxSurge`**, so it takes precedence for *any* `maxSurge` value, not only `0`. It applies even when `maxSurge` is non-zero: a workload can keep `maxSurge: 2` (or any value) for its rollout strategy and still have the eviction autoscaler use a different, workload-tuned batch size during a drain. `maxSurge: 0` is the motivating scenario (SDP guidance mandates it for rollouts), but gating the override on `maxSurge == 0` would re-couple the two knobs we're deliberately separating. When no override annotation is set, we fall straight through to `maxSurge` exactly as today.

internal/controller/evictionautoscaler_controller.go
NamanBhat: Done ? reworked to your proposal: the per-workload `surge-override` annotation is **gone**, replaced by a fleet-wide controller setting (`ae52d7f`).

**Behaviour:** when a workload's `maxSurge` resolves to 0 ? an explicit `maxSurge: 0` **or** a `Recreate` strategy ? and `OVERRIDE_ZERO_MAXSURGE=true`, the controller surges by `SURGE_OVERRIDE_MAX_PERCENT` of `minReplicas` instead of returning `errMaxSurgeZero`. Workloads with a non-zero `maxSurge` are untouched and keep using their own `maxSurge`. Off by default ? today's degrade behaviour preserved. As you noted, the surge stays demand-driven (`minReplicas + displaced`, capped at the fleet percent), so larger drains proceed in waves.

- Toggle: `OVERRIDE_ZERO_MAXSURGE` (bool, strict-validated, default false).
- Percent: reused `SURGE_OVERRIDE_MAX_PERCENT` (default 25; set e.g. `10` at extension-install time).
- Removed `SurgeOverrideAnnotationKey`, `surgeOverrideValue`, `capOverrideSurge`. Kept `surgeFromValue` + tests (they back the % math).
- `DrainSurgeOverride` event still emitted (behind `ENABLE_EVENT_RECORDING`) when the override fires.

Unit + envtest coverage updated (incl. a Recreate-strategy case and the flag-off degrade path); full controller suite passes (97 specs). README updated to describe the fleet-wide setting.

NamanBhat:
**Update (superseding my earlier replies in this PR):** we've gone with the fleet-wide approach, so my earlier argument for a per-workload, decoupled-from-`maxSurge` override no longer stands. The drain surge is now a controller-level `OVERRIDE_ZERO_MAXSURGE` setting that applies **only when `maxSurge` resolves to 0** (explicit `maxSurge: 0` or `Recreate`); non-zero `maxSurge` workloads are unaffected, and there is no per-workload annotation. Details and code are in the resolved thread above (`ae52d7f`).

skuchipudi295:
### Code review

No issues found above the high-confidence threshold. Checked for bugs, historical context, prior PR feedback, and code comment compliance.

**Non-blocking note (scored 75/100):** `ENABLE_EVENT_RECORDING` is parsed in `cmd/main.go` and documented in README, but is not wired through the Helm chart (`deployment.yaml` / `values.yaml`). `ZERO_SURGE_OVERRIDE` is correctly wired. Helm users cannot enable event recording without manually patching the deployment.

https://github.com/Azure/eviction-autoscaler/blob/0f8519ad2b8aff255ffb2a879f3cb24229727a3e/helm/eviction-autoscaler/templates/deployment.yaml#L42-L45

Reviewed at commit 0f8519ad2b8aff255ffb2a879f3cb24229727a3e.

cmd/main.go
```diff
@@ -195,11 +196,51 @@ func main() {
...
+	switch v := os.Getenv("ENABLE_EVENT_RECORDING"); v {
+	case "", "false":
+		eventRecordingEnabled = false
+	case "true":
+		eventRecordingEnabled = true
+	default:
+		setupLog.Error(fmt.Errorf("invalid value %q for ENABLE_EVENT_RECORDING (must be \"true\" or \"false\")", v),
+			"Failed to parse ENABLE_EVENT_RECORDING env variable")
+		os.Exit(1)
+	}
```
skuchipudi295: do we really need this anymore?

internal/controller/evictionautoscaler_controller.go
```diff
@@ -178,11 +191,11 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
 	// surgeTarget = minReplicas + displaced, capped at minReplicas + maxSurge.
 	// If displaced == 0 the formula yields minReplicas, so no scale-up fires and
 	// we fall through to the cooldown/scale-down path — which is correct.
-	maxSurgeTarget, surgeErr := calculateSurge(ctx, target, EvictionAutoScaler.Status.MinReplicas)
+	maxSurgeTarget, surgeErr := calculateSurge(ctx, target, EvictionAutoScaler.Status.MinReplicas, r.ZeroSurgeOverride)
 	if surgeErr != nil {
 		switch {
 		case errors.Is(surgeErr, errMaxSurgeZero):
```
skuchipudi295, line 204: this should happen when maxSurge = 0 & override_surge left empty or unset?

skuchipudi295: comment is confusing

internal/controller/evictionautoscaler_controller.go
```diff
@@ -214,6 +227,20 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		maxSurge := target.GetMaxSurge()
+		if r.ZeroSurgeOverride != nil && isZeroSurge(maxSurge) {
+			msg := fmt.Sprintf("zero-maxSurge override surging %s/%s during drain (rollout maxSurge %q)",
+				target.Obj().GetNamespace(), target.Obj().GetName(), maxSurge.String())
+			logger.Info(msg, "surgeTarget", surgeTarget)
+			if r.Recorder != nil {
+				r.Recorder.Event(target.Obj(), corev1.EventTypeNormal, "DrainSurgeOverride", msg)
+			}
+		}
+
```
skuchipudi295, line 241: I dont think event reporting is necessary here.. logs are enough.. add metrics instead on how many maxSurge:0 deployments are present in a cluster

internal/controller/evictionautoscaler_controller.go
NamanBhat: Confirmed — the `errMaxSurgeZero` degrade fires exactly when `maxSurge` resolves to 0 **and** `ZeroSurgeOverride` is nil (env unset/empty, or a value that resolves to 0, which `ParseZeroSurgeOverride` normalizes to nil). If `maxSurge` is non-zero, or the override is a positive int/percent, this path isn't taken.

Also rewrote the confusing comment (38c275e): the inline comment now says "maxSurge resolves to 0 and no ZeroSurgeOverride set — nothing to surge, degrade" (dropped the misleading "zero override" clause), and the comment above the call explains `calculateSurge` returns the ceiling (`minReplicas + maxSurge`, or `minReplicas + override` when maxSurge=0) with `surgeTarget` below being demand-driven and clamped to it.

internal/controller/evictionautoscaler_controller.go
NamanBhat: Done (38c275e). Removed the `DrainSurgeOverride` event (kept the structured log), and added a gauge metric `eviction_autoscaler_zero_maxsurge_workloads{namespace,name}` — set to 1/0 per workload each reconcile, so the series sum is the count of maxSurge:0 workloads currently in the cluster (independent of the override). It's set right after the target fetch, so it reflects steady state, not only during drains.

cmd/main.go
NamanBhat: Removed (38c275e). Since the event is gone, `ENABLE_EVENT_RECORDING` (which I'd added only to gate it) and its wiring are no longer needed — dropped the env parsing and the recorder wiring in `main.go`. The pre-existing `Recorder` field on the reconciler is left untouched. Structured logs + the new metric are the signal now.

NamanBhat:
Re the automated review's non-blocking note (ENABLE_EVENT_RECORDING not wired through the Helm chart): that env var has been **removed entirely** in 38c275e, per the inline review. The `DrainSurgeOverride` event is dropped in favour of a structured log plus a new gauge metric `eviction_autoscaler_zero_maxsurge_workloads{namespace,name}`, so there's no longer an event-recording toggle to wire through the chart. `ZERO_SURGE_OVERRIDE` remains correctly wired via `values.yaml` + `deployment.yaml`.

skuchipudi295:
### Code review

_(Re-review against 38c275e)_

skuchipudi295 (approved):

skuchipudi295:
### Code review

No issues found. Checked for bugs, historical context, prior PR feedback, and code comment compliance.

Reviewed at commit e2c13f64727f5b9b0cba3ec051d3a4bbd67ccf55.

Issue #10, opened 2025-02-26
What about MaxUnavailble?


paulgmiller:
Surging helps minavailable if there's a single pod stuck unready but with low maxunavailable (1% for example) once you have a single unready replica you could be stuck. For example a 2 replica deployment with maxunavailable 1% is blocked with one unready pod but until you surge it to 100 pods you're still stuck.

Some options.

Mutate pdbs from maxunavailable to minavailble.
Dynamically increase maxunavailable when we surge and then restore it (hard for percent)
Increase not by the max surge but by whatever number is needed
Delete unready pods after we surge in pods if they become ready.
Some tradeoffs to consider. Need some data on how often people set low maxunavailable.

paulgmiller:
Comments just on pr description:
>>maxUnavailable: 1 (a common safe-deployment requirement). As wikipedia would say citation needed. Do you have links to docs that promot this as a requirement. We were always curious who would use it and why so having some justifcation would help the authors of this going forward #10

I don't see how min available as an absolute creates the same problem unless they're setting min available higher. Can you give an example. Is F usually higher than what the user manually sets? when would it be lower.  Is it a bad pdb where minavailable is higher than replicas? 


>>Per wave 
Need to deifine what you mean per wave. Assumign a wave is a set of drain operations 

Perhaps another options is to not do this just on drain but have the pdb reconciler have an option to treat minavailable < replicas as a error and just fix the pdb (or block). Could do a similiar thing with maxunavailble: 1

PR #148, opened 2026-07-18
feat: pin an absolute PDB floor during drains (opt-in) so surge becomes effective

NamanBhat:
## What / why

With an **absolute** PodDisruptionBudget — `maxUnavailable: 1` (a common safe-deployment requirement) or an absolute `minAvailable` — the eviction autoscaler's replica **surge does not raise `DisruptionsAllowed`**: the PDB's implied floor rises with the surged replica count, so once the surge pods are Ready `DisruptionsAllowed` is still `1` and the drain crawls **one pod per wave**. The surge is absorbed by the PDB for exactly the workloads that most need disruption headroom.

## Change

Add an **opt-in** PDB-floor mutation. During a drain the controller temporarily pins the target PDB to an absolute `minAvailable = F` (F = the partner's required-healthy count captured *pre-surge*), so surge headroom converts into real `DisruptionsAllowed`; the partner's original spec is restored when the drain finishes. This preserves the partner's absolute disruption tolerance exactly — it changes the *shape* of the guarantee (relative → absolute, so it stops tracking surged replicas), never the *number* of pods required healthy.

**Lifecycle**
- **Floor F** captured once from `pdb.Status.DesiredHealthy` at the pre-surge `DisruptionsAllowed==0` moment, only when the target is at its baseline replica count (never recomputed against surged replicas). Recorded on the PDB (`.../pinned-floor`) **and** mirrored on the CR `.status.pinnedPDBFloor` for durability against either write being lost.
- **Snapshot S** of the partner spec stored on the PDB (`.../original-pdb-spec`); re-captured if a partner overwrites the PDB mid-drain (last-writer-intent).
- **Re-mutation guard** re-pins F while the drain is still blocking (`DisruptionsAllowed==0`) if the PDB no longer carries the floor.
- **Restore** on scale-down, on the idle path, via a **finalizer** on CR deletion, and via a **stale-window backstop** (`PDB_MUTATION_STALE_WINDOW`, default `2h`).

**Opt-in (both required)** so the feature never rewrites a user-authored PDB without explicit consent:
1. `ENABLE_PDB_FLOOR_MUTATION` controller env flag (default **off**), and
2. per-namespace annotation `eviction-autoscaler.azure.com/pdb-floor-mutation: "true"`.

## Testing

- `go build ./...`, `go vet ./...` clean.
- Unit tests for the helpers + envtest coverage for the reconcile wiring: pin-on-surge, disabled no-op, namespace-not-opted-in, partner-overwrite re-pin, scale-down restore, CR-deletion finalizer restore, stale backstop, and cleanup when the flag is disabled mid-drain. All specs pass.

## Design note for maintainers (consent granularity)

Consent is modeled at the **namespace** level (mirroring the existing `Filter` namespace-enable annotation). A namespace owner and a PDB author aren't always the same principal, so a namespace-level opt-in consents on behalf of every PDB in the namespace. If you'd prefer strict **per-PDB** opt-in (an annotation checked on the PDB itself), that's a small additive change — happy to switch if you want that as the consent unit before merge.

## Notes

- New CR status field `PinnedPDBFloor *int32` (deepcopy + CRD updated).
- No admission webhook; no permanent partner PDB change; no surge-sizing changes.


internal/controller/evictionautoscaler_controller.go
```diff
@@ -66,6 +68,34 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+			case err == nil:
+				if changed, rerr := restorePDBSpec(pdb); rerr != nil {
+					// Can't restore a snapshot we can't parse — do NOT block CR
+					// deletion on it. Log and proceed to drop the finalizer.
+					logger.Error(rerr, "failed to restore PDB on deletion; removing finalizer anyway to avoid a stuck CR", "pdb", pdb.Name)
+				} else if changed {
+					if uerr := r.Update(ctx, pdb); uerr != nil {
+						return ctrl.Result{}, uerr
+					}
+				}
```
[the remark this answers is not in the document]
NamanBhat: Addressed in 3e8cb1e. This is a deliberate non-blocking-deletion trade-off (blocking the finalizer on a corrupt snapshot would wedge the CR in Terminating forever), so we keep removing the finalizer but now emit an operator-visible Warning event (`PDBFloorRestoreFailed`) on the PDB so the un-restored pin is not silent.

internal/controller/evictionautoscaler_controller.go
```diff
@@ -352,3 +509,94 @@ func calculateSurge(_ context.Context, target Surger, minrepicas int32) (int32,
...
+	if controllerutil.RemoveFinalizer(eas, PDBFloorFinalizer) {
+		if err := r.Update(ctx, eas); err != nil {
+			return err
+		}
+	}
+	// Set after the finalizer Update above, which would otherwise refresh the
+	// status back from the server.
+	eas.Status.PinnedPDBFloor = nil
+	return nil
+}
```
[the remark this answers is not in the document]
NamanBhat: Good catch — fixed in 3e8cb1e. `revertPDBFloor` now detects the "snapshot annotation gone but PDB still carries our pinned floor" case: it emits a Warning event (`PDBFloorRestoreFailed`) and removes our `pinned-floor` marker so we stop claiming the pin, instead of silently dropping all tracking. We cannot reconstruct the partner's original spec once the snapshot is gone, so the (stricter) minAvailable floor is intentionally left in place as the fail-safe direction. Added an envtest for this path.

internal/controller/pdbmutation.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	v, ok := pdb.Annotations[AnnotationPinnedFloor]
+	if !ok {
+		return 0, false
+	}
+	f, err := strconv.Atoi(v)
+	if err != nil {
+		return 0, false
+	}
+	return int32(f), true
```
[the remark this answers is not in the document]
NamanBhat: Fixed in 3e8cb1e — `pinnedFloorFromPDB` now uses `strconv.ParseInt(v, 10, 32)`, which rejects any value outside the int32 range instead of truncating. Added unit tests for overflow and non-numeric values.

cmd/main.go
```diff
@@ -182,9 +182,10 @@ func main() {
 	setupLog.Info("PDB creation configuration", "pdbCreate", pdbCreate)
 
 	if err = (&controllers.EvictionAutoScalerReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
-		Filter: nsfilter,
+		Client:   mgr.GetClient(),
+		Scheme:   mgr.GetScheme(),
+		Filter:   nsfilter,
+		Recorder: mgr.GetEventRecorderFor("eviction-autoscaler"),
```
skuchipudi295: please dont use event recorder.. it is too much noise instead emit metrics so we can configure alerts

cmd/main.go
```diff
@@ -182,9 +182,10 @@ func main() {
 	setupLog.Info("PDB creation configuration", "pdbCreate", pdbCreate)
```
skuchipudi295, line 199: need to log configuration option for pdb mutation here, basically need some validation for it in main.go

config/crd/bases/eviction-autoscaler.azure.com_evictionautoscalers.yaml
```diff
@@ -133,6 +133,16 @@ spec:
               minReplicas:
                 format: int32
                 type: integer
+              pinnedPDBFloor:
```
skuchipudi295, line 136: why a field vs annotation?

skuchipudi295:
Additional requirement: **Event recording should be behind a dedicated, separate controller flag (`ENABLE_EVENT_RECORDING`, default off).**

This is orthogonal to feature flags — it's an installation-time decision about observability verbosity, not about feature behavior. Implementation:

```go
// main.go
eventRecordingEnabled := os.Getenv("ENABLE_EVENT_RECORDING") == "true"

reconciler := &controllers.EvictionAutoScalerReconciler{
    Client: mgr.GetClient(),
    Scheme: mgr.GetScheme(),
    Filter: nsfilter,
}
if eventRecordingEnabled {
    reconciler.Recorder = mgr.GetEventRecorderFor("eviction-autoscaler")
}
```

Then the existing `if r.Recorder != nil` guard throughout the controller naturally gates all event emission — no per-feature event checks needed.

**Why separate from feature flags:**
- Single knob for all event noise across all features (surge override, PDB mutation, future features)
- Operators who know their fleet is large enough that events are noise can disable globally at install time
- Feature flags control *behavior*; this flag controls *observability verbosity* — orthogonal concerns
- Default off: structured logs + metrics are the primary path; events are opt-in for operators who actually use `kubectl describe`

Validate this the same way as other env vars — fail fast on invalid values.

Also applies to #147 — same pattern across both PRs.

cmd/main.go
NamanBhat: Done. `ENABLE_PDB_FLOOR_MUTATION` is now parsed, validated and logged in `main.go` — a strict `true`/`false`/empty switch with fail-fast, plus a startup log line — and threaded onto the reconciler as `PDBFloorMutationEnabled`, matching the `PDB_CREATE` pattern. Previously it was resolved silently in a package var that ignored parse errors (a typo like `yes` → silently off); now a misconfiguration fails fast at startup. Done in `de8c57c`.

cmd/main.go
NamanBhat: Done. The restore-failure now increments a Prometheus counter `eviction_autoscaler_pdb_floor_restore_failures_total` (labels `namespace`, `pdb_name`, `reason` = `deletion_corrupt_snapshot` / `snapshot_missing`) at both restore-failure sites — that's the alertable signal. And per your other comment, I gated events behind a dedicated `ENABLE_EVENT_RECORDING` flag (default off): the `Recorder` is only wired when it's on, so the `PDBFloorRestoreFailed` event no longer fires by default (no noise) — metrics/logs are the primary path. Added a metric-increment test + README docs. Done in `de8c57c`.

config/crd/bases/eviction-autoscaler.azure.com_evictionautoscalers.yaml
NamanBhat: `pinnedPDBFloor` lives on our own EvictionAutoScaler `.status` because it's controller-owned **observed state** — which is exactly what `.status` is for: it's typed (`*int32`, schema-validated by the API server), clearly controller-managed (status subresource), and surfaces where operators expect observed state. An annotation would be stringly-typed and hand-editable. The reason the PR uses an *annotation* for the PDB snapshot (`original-pdb-spec`) is the opposite case — the PDB is a foreign object whose schema we don't own, so we can't add a status field there. So the split is deliberate: status field on our CR (state we own), annotation on the PDB (foreign object). Happy to revisit if you'd prefer to keep it off the CRD surface.

internal/controller/pdbmutation.go
```diff
@@ -0,0 +1,219 @@
...
+
+	// envStaleMutationWindow overrides the stale window without rebuilding. Value
+	// must be a Go duration string ("30m", "2h", ...).
+	envStaleMutationWindow = "PDB_MUTATION_STALE_WINDOW"
+
+	// defaultStaleMutationWindow bounds how long a PDB may stay mutated before the
+	// backstop restores it unconditionally. Generous by default because a healthy
+	// drain reverts on its own well within it; this only catches orphaned
+	// mutations (e.g. the CR deleted without the finalizer running).
+	defaultStaleMutationWindow = 2 * time.Hour
```
skuchipudi295: 2 hours? do you expect drains to be that long?

internal/controller/pdbmutation.go
```diff
@@ -0,0 +1,219 @@
...
+// mutation. The reconcile loop guarantees this via pdbCarriesFloor.
+func snapshotPDBSpec(pdb *policyv1.PodDisruptionBudget) error {
+	specBytes, err := json.Marshal(pdb.Spec)
+	if err != nil {
+		return fmt.Errorf("snapshotPDBSpec: marshal spec: %w", err)
+	}
+	if pdb.Annotations == nil {
+		pdb.Annotations = map[string]string{}
+	}
+	pdb.Annotations[AnnotationOriginalPDBSpec] = string(specBytes)
```
skuchipudi295, line 165: do we need snapshot entire specs here?

internal/controller/evictionautoscaler_controller.go
```diff
@@ -66,6 +75,37 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+			case !apierrors.IsNotFound(err):
+				return ctrl.Result{}, err
+			}
+			controllerutil.RemoveFinalizer(EvictionAutoScaler, PDBFloorFinalizer)
+			if err := r.Update(ctx, EvictionAutoScaler); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+		return ctrl.Result{}, nil
+	}
```
skuchipudi295, line 117: this should be after namespace enabled check

skuchipudi295:
### Code review

The PDB-floor mutation logic is well-structured:
- Floor F is captured exactly once at the right moment (pre-surge, baseline replica count), never recomputed
- Dual-write durability (CR status + PDB annotation) with proper fallback chain
- Re-mutation guard correctly detects partner overwrites via `pdbCarriesFloor`
- Snapshot-lost case is handled safely (leaves stricter floor in place + metric)
- Finalizer removal does not block on corrupt snapshots (correct trade-off to avoid stuck Terminating CRs)
- `ParseInt` with bitSize 32 properly rejects overflow from user-editable annotations
- Stale-window backstop defaults to safe direction (restore) for missing/malformed timestamps

internal/controller/pdbmutation.go
NamanBhat: Not the expected drain length — normal drain completion reverts the mutation immediately via the surge-revert path (when the drain finishes and `DisruptionsAllowed` reopens, `revertPDBFloor` restores the partner's spec). This `staleMutationWindow` is purely an **orphan backstop**: it only ever fires for a mutation left behind because the CR was deleted/force-removed without the finalizer running. Since the healthy path never relies on it, I've kept the default generous (and it's env-tunable via `PDB_MUTATION_STALE_WINDOW`) so it never races a legitimately long drain — if it ever did fire early, it would just restore→re-pin on the next reconcile.

internal/controller/pdbmutation.go
NamanBhat: Good point — we don't need the whole spec. We only ever mutate `minAvailable`/`maxUnavailable`, so I've optimized the snapshot to store just those two disruption fields instead of the entire `PodDisruptionBudgetSpec`. Besides being smaller, restore now writes back only the fields we manage, so any concurrent partner edits to other fields (e.g. `selector`) are preserved rather than overwritten by the snapshot on revert. Added a unit test covering the selector-preservation case. (a90e20a)

internal/controller/evictionautoscaler_controller.go
NamanBhat: I'd keep the deletion/finalizer-restore block before the namespace-enabled check. If it runs *after* the gate, this failure mode appears: a namespace has the feature on, a drain pins a PDB and the CR gets the `PDBFloorFinalizer`; the namespace opt-in is later removed (or the master switch flipped off); then the CR is deleted. With deletion-handling after the gate, the `!isEnabled` early-return fires first, so the finalizer branch never runs → the finalizer is never removed → the CR is stuck in `Terminating` forever and the partner's PDB is left pinned. Finalizer cleanup has to run regardless of current enablement (it's undoing state we already created), which is the standard controller-runtime idiom — handle `DeletionTimestamp` + finalizers at the top of `Reconcile`. It's already cheaply guarded by `ContainsFinalizer`, so it's a no-op for any CR we never touched.

skuchipudi295:
### Code review

Found 2 issues:

1. Floor capture can be permanently skipped when `DesiredHealthy` is stale at first surge

In `ensurePDBFloor`, the floor is captured from `pdb.Status.DesiredHealthy`. But `DesiredHealthy` is computed asynchronously by the PDB controller. If it hasn't been reconciled yet (returns 0), the `floor <= 0` guard skips the pin silently. The surge then proceeds without a pinned floor. On the next reconcile, replicas are already surged so `atBaseline` is false, and the code refuses to capture ("an overlapping drain"). The floor is never pinned for that drain cycle, defeating the feature. Consider retrying with a requeue instead of silently skipping when `DesiredHealthy == 0` and the PDB exists.

https://github.com/Azure/eviction-autoscaler/blob/64051d925ff4bf8f281f407b1483565a4c465c5b/internal/controller/evictionautoscaler_controller.go#L400-L410

2. `PDB_MUTATION_STALE_WINDOW` silently falls back to default on invalid input, unlike the other env vars

`ENABLE_PDB_FLOOR_MUTATION` and `ENABLE_EVENT_RECORDING` are strictly validated in `main.go` with `os.Exit(1)` on bad values. But `PDB_MUTATION_STALE_WINDOW` (in `pdbmutation.go`) silently falls back to `2h` if the value is unparseable or negative. A typo like `PDB_MUTATION_STALE_WINDOW=2hr` would be silently ignored. Consider moving the parse+validation to `main.go` with the same fail-fast pattern, or at minimum logging a warning.

https://github.com/Azure/eviction-autoscaler/blob/64051d925ff4bf8f281f407b1483565a4c465c5b/internal/controller/pdbmutation.go#L68-L74

_(Review against 64051d9)_

skuchipudi295:
### Code review

Found 4 issues:

1. Floor capture can be permanently skipped for a drain cycle when `DesiredHealthy` is stale or 0

In `ensurePDBFloor`, when no existing floor is found (first capture), `floor = pdb.Status.DesiredHealthy` is read. If the PDB controller hasn't reconciled yet (e.g. PDB just created, or apiserver lag), `DesiredHealthy` is 0. The `floor <= 0` guard silently returns `(0, false, nil)` — no error, no requeue hint. The surge then proceeds without a pinned floor. On the next reconcile, replicas are already surged, so `atBaseline` is false and the code refuses to capture ("refuse to capture an inflated floor"). The floor is never pinned for that entire drain cycle, silently defeating the feature. Consider returning a `RequeueAfter` or explicit error when `DesiredHealthy == 0` and the PDB exists and has a non-zero selector match, so the controller retries before surging.

https://github.com/Azure/eviction-autoscaler/blob/64051d925ff4bf8f281f407b1483565a4c465c5b/internal/controller/evictionautoscaler_controller.go#L540-L550

2. `PDB_MUTATION_STALE_WINDOW` silently falls back to default on invalid input — inconsistent with the strict validation of other env vars

`ENABLE_PDB_FLOOR_MUTATION` and `ENABLE_EVENT_RECORDING` are strictly validated in `main.go` with `os.Exit(1)` on bad values. But `PDB_MUTATION_STALE_WINDOW` in `pdbmutation.go` uses a package-level `var` init that silently falls back to `2h` if the value is unparseable or negative. A typo like `PDB_MUTATION_STALE_WINDOW=2hr` (should be `2h`) would be silently ignored — the operator thinks they set a 2-hour window but actually got the default 2h (lucky coincidence here, but `30min` vs `30m` would be silent and wrong). Consider moving the parse to `main.go` with the same fail-fast pattern.

https://github.com/Azure/eviction-autoscaler/blob/64051d925ff4bf8f281f407b1483565a4c465c5b/internal/controller/pdbmutation.go#L68-L74

3. Re-mutation guard passes `atBaseline=true` unconditionally, bypassing the inflated-floor protection

In the re-mutation guard (controller line ~240), when `floorKnown && pdb.Status.DisruptionsAllowed == 0 && !pdbCarriesFloor(pdb, pinnedFloor)`, the code calls `r.ensurePDBFloor(ctx, EvictionAutoScaler, pdb, true)` with `atBaseline=true` hardcoded. This is fine for re-pinning an existing floor, but `ensurePDBFloor` has a `default` branch that captures a *new* floor from `DesiredHealthy` when no floor exists yet. If both the PDB annotation and the CR status have lost the floor (e.g. both were stripped), this path would capture `DesiredHealthy` at an already-surged replica count, defeating the inflated-floor guard. The `atBaseline` parameter should reflect actual state, or the re-mutation guard should only call `ensurePDBFloor` when `floorKnown` is true (which it is, by the guard condition — so this is safe in practice, but the `true` literal is misleading and fragile if the guard condition is ever relaxed).

https://github.com/Azure/eviction-autoscaler/blob/64051d925ff4bf8f281f407b1483565a4c465c5b/internal/controller/evictionautoscaler_controller.go#L240-L250

4. Multiple sequential `r.Update` calls in `revertPDBFloor` without conflict retry — PDB mutation left in inconsistent state on partial failure

`revertPDBFloor` does up to 3 sequential `r.Update` calls: (1) PDB restore, (2) PDB annotation cleanup (snapshot-missing path), (3) CR finalizer removal. If call (1) succeeds but (3) fails (e.g. conflict on the CR), the PDB is restored but the finalizer remains. On the next reconcile, the deletion handler will try to restore the PDB again — `restorePDBSpec` returns `(false, nil)` since annotations are already cleared, so the finalizer is removed. This is idempotent and safe. However, in the snapshot-missing path, if the PDB annotation cleanup (2) succeeds but the CR finalizer removal (3) fails, the next reconcile sees no `isMutated` (annotations cleared) and no `PinnedPDBFloor` (cleared in memory but not persisted), but the finalizer is still present — it will be removed on the next reconcile. Again safe but worth documenting the idempotency assumptions.

https://github.com/Azure/eviction-autoscaler/blob/64051d925ff4bf8f281f407b1483565a4c465c5b/internal/controller/evictionautoscaler_controller.go#L577-L610

_(Thorough re-review against 64051d9)_

skuchipudi295:
Issue 1 — Floor F is captured from an asynchronous status field (correctness)

In ensurePDBFloor, the floor is taken from [pdb.Status.DesiredHealthy](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), gated only on [atBaseline = target.GetReplicas() == Status.MinReplicas](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). DesiredHealthy is computed asynchronously by the built-in PDB controller and can lag the current replica count. The [atBaseline](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) check looks at the deployment's replicas, not at whether the PDB status is fresh for that replica count — so if DesiredHealthy still reflects a prior count, the pinned F is off by the lag and is then held constant for the whole drain (never recomputed). Suggest gating on PDB status freshness (e.g. [Status.ObservedGeneration](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)) or deriving F from the snapshotted absolute spec + baseline replicas instead of the async DesiredHealthy.

Issue 2 — Write amplification increases optimistic-concurrency conflict churn (hardening)

A single surge reconcile can issue three writes to two objects: [Update(eas)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) for the finalizer, [Update(pdb)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) for the pin, then [Status().Update(eas)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) for the floor. Each can lose a conflict race with an external PDB/CR writer, and any conflict re-runs the whole reconcile. It does converge, and the dual-write of F onto both the PDB annotation and [Status.PinnedPDBFloor](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is a sound durability choice — but the separate finalizer Update (which re-fetches and discards in-memory state) is avoidable. Consider adding the finalizer in the same Update that pins the PDB, or ordering writes to minimize the window.

Issue 3 — Latent test data race on a package global (hardening)

[staleMutationWindow](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is a mutable package-level var that tests override directly. If any specs that touch it ever run in parallel in the same process, that's a data race. -race is clean today because the specs run serially, but a shared mutable global for test injection is fragile — prefer a reconciler field or a per-test env value.

Issue 4 — Cross-controller contention is well-defended; please lock it in with a test

The dangerous interaction — the surge changing the deployment/HPA and triggering deployment_to_pdb_controller / autoscaler_to_pdb_controller to rewrite [Spec.MinAvailable](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), clobbering the pin and causing the re-mutation guard to re-snapshot a surge-derived value as "original" — is correctly prevented today by three guards: (a) the pin writes only the PDB, so it doesn't trigger the siblings; (b) ApplySurge sets the surge annotation atomically with the [minReplicas](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) change, so autoscaler_to_pdb's isSurgeActiveOnAutoscaler skip always fires; (c) siblings only touch ownedBy PDBs, so user-authored maxUnavailable: 1 PDBs are never rewritten. This is subtle and easy to regress. Please add a regression test asserting that a sibling reconcile during an active pin is a no-op.

NamanBhat:
Thanks for the thorough re-review — addressed in `1473ba6` (fixes) and `20c2668` (lint refactor). Point by point:

**Issue 1 — Floor `F` from async `DesiredHealthy` (correctness).** Fixed. `F` is now derived **synchronously** from the partner's PDB spec at the baseline replica count (`Status.MinReplicas`) via a new `desiredHealthyAt` helper that mirrors the built-in disruption controller (`intstr.GetScaledValueFromIntOrPercent`, roundUp). This removes the dependency on the asynchronously-populated `Status.DesiredHealthy`, so a stale/`0` read can no longer silently skip the pin. Because it's derived from the stable baseline `MinReplicas` (not live replicas), it stays correct even after replicas inflate — which let me **drop the fragile `atBaseline` gate entirely** (this also resolves the `atBaseline=true` literal nit). Added a regression test: with `DesiredHealthy=0`, the floor still pins from spec. *Known limitation, commented inline:* for an autoscaler running above its min, `F` reflects the min rather than the elevated count (outside the primary Deployment/ReplicaSet scope, and not a regression — the old `atBaseline` gate never engaged there either).

**Stale-window validation + mutable global (your Issue 3 / the bots' Issue 2).** Both fixed together: `PDB_MUTATION_STALE_WINDOW` is now parsed **and strictly validated in `main.go`** (fail-fast on malformed/non-positive, matching `ENABLE_PDB_FLOOR_MUTATION`) and threaded onto the reconciler as a `StaleMutationWindow` field. The mutable package global is gone, so there's no shared-state race — tests set the field / pass the window explicitly.

**Issue 2 — write amplification.** Acknowledged; left as-is intentionally. The three writes are on **different objects/subresources** — the finalizer on the CR metadata, the pin on the PDB, and the floor on the CR *status* subresource — which controller-runtime can't collapse into one call. The current order (finalizer -> pin -> status) is deliberate (the finalizer `Update` refreshes from the server first), and it converges correctly on conflict via re-reconcile. Happy to revisit if you'd like a specific reordering.

**Issue 4 — cross-controller contention regression test.** Added. New `autoscaler_to_pdb_pin_test.go` asserts the `AutoscalerToPDB` sibling is a **no-op on a pinned PDB while a surge is active** (the `isSurgeActiveOnAutoscaler` guard), plus a **control** case showing that without the active surge it *would* rewrite `minAvailable` to the autoscaler floor — so the test fails loudly if the guard is ever removed.

**`revertPDBFloor` idempotency (bot Issue 4).** No logic change (already safe); added a doc comment spelling out that each write is idempotent and a partial failure converges on the next reconcile.

**CI / lint (`gocyclo`).** The floor feature had pushed `Reconcile`'s cyclomatic complexity to 55 (over the repo's 30 limit), which was failing the `build` check independently of the review. Extracted the floor lifecycle and surge phases into behavior-preserving helpers (`reconcileFloorOnDeletion`, `restoreStaleFloor`, `restoreFloorIfDrainHandled`, `defendPinnedFloor`, `pinFloorBeforeSurge`, `handleBlockedDrain`, `handleScaleDown`) — `Reconcile` is now 28, no behavior change (`20c2668`).

Build + vet clean; full controller suite passes.

skuchipudi295:
### Code review

Reviewed at commit 20c2668b522af1a9316e5d458712a93d17cf5873.

skuchipudi295:
### Code review

Found 1 issue:

1. **`handleScaleDown` ordering: revert surge before PDB restore can briefly stall new drains**

The scale-down path reverts the deployment (`RevertSurge`) before restoring the partner PDB (`revertPDBFloor`). If the PDB restore write hits a conflict, the deployment is already scaled back to `minReplicas` but the PDB is still pinned at the old absolute floor (`minAvailable: F`). Until the next reconcile runs `restoreFloorIfDrainHandled`, the cluster sits in a state where the PDB is more restrictive than the partner intended — this blocks new drains/evictions for the duration of one requeue interval.

This converges (fail-safe direction, the next reconcile cleans it up), but worth documenting the transient stall window. Consider either (a) restoring the PDB _before_ reverting the surge, or (b) adding a comment in `handleScaleDown` acknowledging the ordering tradeoff and why fail-safe (over-restrictive) is preferred over fail-open (under-replicated + permissive PDB).

https://github.com/Azure/eviction-autoscaler/blob/d1b4070a96a1bfe5ea22285e2a4aca33b61f25b3/internal/controller/evictionautoscaler_controller.go#L607-L622

skuchipudi295:
### Follow-up: `defendPinnedFloor` does not distinguish partner intent from accidental overwrite

`defendPinnedFloor` unconditionally re-pins the floor whenever `pdbCarriesFloor(pdb, floor)` returns false. It does not check whether the partner's new spec is **more** or **less** restrictive than the pinned floor.

**Example:** Floor is pinned at `minAvailable: 4`. A partner (or operator) intentionally patches the PDB to `minAvailable: 6` mid-drain because they need stricter protection. On the next reconcile:

1. `pdbCarriesFloor(pdb, 4)` → false (it's 6, not 4)
2. `defendPinnedFloor` fires, snapshots `{"minAvailable": 6}`, re-pins to `minAvailable: 4`
3. We just **weakened** the PDB below what the partner explicitly requested

This is the wrong direction — the controller is supposed to be a safety backstop, not override a partner who is asking for **more** protection.

| Partner changes to | vs. floor (4) | Current behavior | Correct behavior |
|---|---|---|---|
| `minAvailable: 2` | Less restrictive | Re-pin to 4 | Re-pin to 4 ✅ (defend drain) |
| `minAvailable: 6` | More restrictive | Re-pin to 4 ❌ | Respect it (or raise floor to 6) |
| `maxUnavailable: N` | Different shape | Re-pin to 4 | Debatable — depends on resolved desiredHealthy |

**Suggested fix:** Before re-pinning, compute `desiredHealthyAt(pdb.Spec, currentReplicas)` for the partner's new spec. If it resolves to `>= floor`, the partner is being stricter than us — skip the re-pin (or adopt their higher floor). Only re-pin when their change would lower protection below what the drain needs.

skuchipudi295:
### Follow-up: PDB floor pinning is unnecessary for absolute `minAvailable`

The PR pins the floor for all PDB types, but an absolute `minAvailable` already acts as a fixed floor — surge naturally increases `DisruptionsAllowed` without any mutation needed.

**How `desiredHealthy` behaves with surge (5→7 pods):**

| PDB type | desiredHealthy formula | DA before surge | DA after surge | Pin needed? |
|---|---|---|---|---|
| `minAvailable: 4` (absolute) | 4 (fixed) | 5-4=1 | 7-4=3 | **No** — already works |
| `minAvailable: 80%` (percentage) | ceil(total×80%) | 5-4=1 | 7-6=1 | **Yes** — floor scales with pods |
| `maxUnavailable: 1` (absolute) | total-1 | 5-4=1 | 7-6=1 | **Yes** — floor scales with pods |
| `maxUnavailable: 20%` (percentage) | total-ceil(total×20%) | 5-4=1 | 7-2=5, DA=2 | **Partially** — helps but less than full |

`ensurePDBFloor` / `pinFloorBeforeSurge` should skip the mutation when the PDB already uses an absolute `minAvailable` — it is a fixed floor by definition and surge alone gives us the DA headroom we need. Pinning it is a no-op at best, and at worst it adds unnecessary annotation churn, finalizer lifecycle, and the partner-overwrite fight described in the previous comment.

The pin is only needed when `desiredHealthy` would scale with replica count: `maxUnavailable` (absolute or percentage) and percentage `minAvailable`.

NamanBhat:
Thanks @skuchipudi295 for the round-2 pass. All three are intentional behavior — details below, and I pushed `abf07fd` with clarifying code comments + a new test (no functional change).

---

**1. `handleScaleDown` ordering: revert surge before PDB restore**

Deliberate fail-safe, now documented in a code comment. We scale the deployment back to `MinReplicas` **before** restoring the partner PDB: if the PDB restore then fails, we requeue with the deployment already at baseline but the floor still pinned — briefly over-restrictive (a brand-new drain may cost one extra requeue) and it converges next reconcile. The reverse order risks the opposite momentary state — a permissive PDB over an under-replicated deployment (fail-open) — so we prefer the over-restrictive failure mode.

---

**2. `defendPinnedFloor` does not distinguish partner intent (direction check)**

Intentional. Once a drain is in flight we hold the floor `F` (captured once at drain start as the pre-surge `desiredHealthy`) constant for the whole drain and defend it against **any** partner edit, so a mid-drain PDB change can't perturb an already-authorized drain wave.

The partner's new intent is **not** discarded: on every reconcile where they've moved the PDB off our pin, `ensurePDBFloor` calls `snapshotPDBSpec` **before** re-pinning, which overwrites the restore snapshot with their *current* spec. So a mid-drain tighten `minAvailable 4->6` is snapshotted, we keep serving at `F`, and `revertPDBFloor` restores their latest spec (`6`) the moment the drain completes — deferred, not lost. And `F` is the pre-surge `desiredHealthy`, so availability never drops below what existed before we started; the pin is bounded by the drain lifetime + stale-window backstop.

---

**3. Floor pinning is unnecessary for absolute `minAvailable`**

I opted to **keep** pinning absolute-`minAvailable` PDBs for a consistent mid-drain contract, and the churn concern is already handled by the re-mutation guard:

- When the PDB already carries the floor (`minAvailable: 4`, floor `4`), `pdbCarriesFloor` is `true`, so `ensurePDBFloor` **short-circuits the snapshot + rewrite** — the PDB spec is left byte-for-byte untouched, no annotations added. The only always-on cost is the finalizer + `Status.PinnedPDBFloor` on **our CR**, not the partner's PDB.
- The PDB is only rewritten **if** the partner moves it off the floor mid-drain — exactly the case we want to defend (same defer-to-restore semantics as the `maxUnavailable`/percentage paths). Skipping the pin here would make absolute-`minAvailable` the one shape where a mid-drain loosening could over-disrupt below the drain-start floor.

Added an integration test (`claims the floor without churning the spec for an absolute-minAvailable PDB`) asserting exactly this: spec + annotations untouched, but finalizer + persisted floor present.

---

I'll also rebase onto latest `main` (now that #147 has merged) and re-run the suite.

paulgmiller:
Seens related to https://github.com/Azure/eviction-autoscaler/issues/10 taking a look

api/v1/evictionautoscaler_types.go
```diff
@@ -24,6 +24,13 @@ type EvictionAutoScalerStatus struct {
 	MinReplicas      int32              `json:"minReplicas"`            // Minimum number of replicas to maintain
 	TargetGeneration int64              `json:"deploymentGeneration"`   // generation (spec hash) of deployment or statefulse
 	Conditions       []metav1.Condition `json:"conditions,omitempty"`
+	// PinnedPDBFloor is the absolute minAvailable value the controller has pinned
+	// onto the target's PDB for the duration of an active drain. It is captured
+	// once (from the PDB's DesiredHealthy at the pre-surge moment) and held
+	// constant until the drain completes, so it is immune to surge inflation and
+	// survives a partner stripping the PDB's snapshot annotations. Nil when no
+	// PDB-floor mutation is active.
+	PinnedPDBFloor *int32 `json:"pinnedPDBFloor,omitempty"`
```
paulgmiller, line 33: not sure this is the right place for this. This are usually thing generated from a pdb but now this points backwwards to things that will change the pdb. Will it ever differ from min repilicas or could we just reuse that?

cmd/main.go
```diff
@@ -195,6 +197,60 @@ func main() {
...
 
+	// Parse ENABLE_PDB_FLOOR_MUTATION environment variable (defaults to false if not set).
+	// Fleet-wide master opt-in for the PDB-floor mutation feature; when off, the
+	// controller never pins/mutates a partner PDB's minAvailable floor. Strictly
+	// validated to "true"/"false"/empty so a typo fails fast at startup rather than
+	// silently disabling the feature.
+	pdbFloorMutationEnabled := false
+	switch v := os.Getenv("ENABLE_PDB_FLOOR_MUTATION"); v {
+	case "", "false":
```
paulgmiller: strconv.ParseBool alos nice here.

cmd/main.go
```diff
@@ -195,6 +197,60 @@ func main() {
...
+		if err != nil || d <= 0 {
+			setupLog.Error(fmt.Errorf("invalid value %q for PDB_MUTATION_STALE_WINDOW (must be a positive Go duration, e.g. \"2h\")", v),
+				"Failed to parse PDB_MUTATION_STALE_WINDOW env variable")
+			os.Exit(1)
+		}
+		staleMutationWindow = d
+	}
+	setupLog.Info("PDB mutation stale window configuration", "staleMutationWindow", staleMutationWindow)
+
+	// Parse ENABLE_EVENT_RECORDING environment variable (defaults to false if not set).
```
paulgmiller: seems like it shoud be a sepertate pr?

paulgmiller: were we not actually eventing before?

paulgmiller (changes requested):
diff is quite long want to settle commetns abotu why we're handling maxunavailable first before we introrduce this much code. 

WHy can't we just reject or mutate maxunavalble permenatnly

skuchipudi295:
@NamanBhat this should be honored by eviction autoscaler and this pr doesnt seem to do that


https://github.com/Azure/eviction-autoscaler/blob/main/README.md#pdb-ownership-and-lifecycle-management

skuchipudi295 (changes requested):

cmd/main.go
skuchipudi295: nope.

skuchipudi295:
@NamanBhat please move eventing to separate pr

cmd/main.go
NamanBhat: Good point. This was originally a strict `switch` because Sushma earlier asked for validation + fail-fast on a bad value. But you're right that `strconv.ParseBool` reads cleaner, and it's what the sibling env vars (`ENABLED_BY_DEFAULT`, `PDB_CREATE`) already use — so I've switched to it for consistency. It still fails fast (`os.Exit(1)`) on an invalid value, so the validation is preserved. Done in `cc93996`.

cmd/main.go
NamanBhat: Good call, Paul — you're right on both counts. We weren't eventing before; the event recorder came in during an earlier review round on this PR, and it's orthogonal to the PDB-floor mutation. I've removed `ENABLE_EVENT_RECORDING` and the event emission from this PR in `8010190` — the restore-failure cases are still surfaced via the `eviction_autoscaler_pdb_floor_restore_failures_total` metric and structured logs. I'll bring the Kubernetes-events piece back as its own separate PR.

api/v1/evictionautoscaler_types.go
NamanBhat: Great questions, thanks for digging into this.

On reusing `MinReplicas` — I looked at that, but they end up being different values, so unfortunately it wouldn't quite work. `PinnedPDBFloor` is the PDB's required-healthy count (its `desiredHealthy`), whereas `MinReplicas` is the deployment's replica floor. For example, with `maxUnavailable: 1` at 5 replicas the floor is **4** while MinReplicas is **5**; similarly `minAvailable: 80%` at 5 replicas gives a floor of **4** vs MinReplicas **5**. They only line up for certain PDB shapes, so reusing `MinReplicas` would risk pinning the wrong floor.

On the placement — that's a fair observation, and I agree it's a slightly unusual shape: normally a Status field reflects state observed *from* the PDB, whereas here it's used to *drive* a change to it (the "points backwards" you mentioned). We landed on `Status` deliberately, since it's controller-observed state that's captured once at drain start, held constant, and needs to survive a partner stripping the PDB's snapshot annotations — so it acts as our durable recovery source of truth for the floor (and it's not partner-facing config, which would belong on Spec).

NamanBhat:
> diff is quite long; want to settle comments about why we're handling maxUnavailable first before we introduce this much code. Why can't we just reject or mutate maxUnavailable permanently?

Thanks for the careful review, Paul — great question, let me walk through the why.

The key thing is that surge alone doesn't help drain agility for most PDBs. Whenever the required-healthy count is **relative to the replica count** — `maxUnavailable` (absolute or %), and also percentage `minAvailable` — the floor rises right along with the replicas we add, so `disruptionsAllowed` stays flat and the surge is absorbed. The one exception is an **absolute `minAvailable`**, which is fixed. So for every replica-relative PDB, the floor has to be held at its pre-surge value during the drain — i.e. pinned to an absolute value temporarily.

On the two shortcuts: rejecting `maxUnavailable` would exclude the common shape these workloads actually use, and permanently converting it to `minAvailable` isn't safe — it's partner-owned intent, and the two forms diverge as the replica count moves (e.g. an HPA/KEDA scaling the workload up or down), so a permanent pin would mis-protect the workload outside of drains. That's why we pin only for the drain and then restore the exact original — most of the "extra" code is just the reversibility safety net (snapshot, finalizer, stale-window backstop).

On diff size — fair, and I've already pulled the orthogonal `ENABLE_EVENT_RECORDING` into its own PR to trim it.

NamanBhat:
> this should be honored by eviction autoscaler and this pr doesnt seem to do that — PDB Ownership and Lifecycle Management

Good point, and you're right that EAS's contract is to only modify PDBs it owns. This PR keeps that contract intact **by default**.

The PDB-floor mutation is off by default and **double-gated**: it only runs when an operator sets the `ENABLE_PDB_FLOOR_MUTATION` install flag (default `false`) **and** a namespace carries the per-namespace opt-in annotation (`eviction-autoscaler.azure.com/pdb-floor-mutation: "true"`). So with the feature disabled — the default — nothing here touches a user-owned PDB, and the ownership/lifecycle contract stands exactly as documented.

Only when the feature is explicitly enabled does EAS **temporarily** pin a user-owned PDB's floor for the duration of an active drain — and it's transient and self-reverting: the original spec is snapshotted and restored on completion, with a stale-window backstop as a safety net. It never takes ownership; it's a bounded, opt-in, reversible operation, distinct from the lifecycle management the `ownedBy` contract governs.

I've documented this explicitly as an opt-in exception in the "PDB Ownership and Lifecycle Management" section you linked (commit `de5fbf4`), so the contract doc itself now calls it out.

internal/controller/evictionautoscaler_controller.go
```diff
@@ -197,16 +223,34 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 	}
 
+	// Re-mutation guard: while a floor is pinned and the drain is still active, keep
+	// defending it against a partner overwriting the PDB mid-drain (the mutated PDB
+	// is the sole gate). The floor is read from the PDB annotation first (survives a
+	// lost CR status write) and falls back to the CR status (survives a partner
+	// stripping the PDB annotations), so the guard fires under either failure. Gated
+	// on the drain still blocking (DisruptionsAllowed==0) so we don't re-pin a PDB
+	// that is about to be reverted this same reconcile.
+	if err := r.defendPinnedFloor(ctx, EvictionAutoScaler, pdb); err != nil {
```
skuchipudi295, line 247: we should not do this.. we should let customer change floor during drain.. this is too restrictive for customer. please remove this.

internal/controller/evictionautoscaler_controller.go
```diff
@@ -229,58 +273,7 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
...
-
-		// Log the scaling action
-		logger.Info(fmt.Sprintf("Scaled up %s %s/%s to %d replicas (via %s)", EvictionAutoScaler.Spec.TargetKind, target.Obj().GetNamespace(), target.Obj().GetName(), surgeTarget, surgeApplier.Name()))
-		logger.Info(fmt.Sprintf("TargetGeneration moving from %d->%d", EvictionAutoScaler.Status.TargetGeneration, target.Obj().GetGeneration()))
-		// Save ResourceVersion to EvictionAutoScaler status this will cause another reconcile.
-		EvictionAutoScaler.Status.TargetGeneration = target.Obj().GetGeneration()
-		//Do not update EvictionAutoScaler.Status.LastEviction because we need to keep reconciling till scale down
-		ready(&EvictionAutoScaler.Status.Conditions, "Reconciled", "eviction with scale up")
-		return ctrl.Result{RequeueAfter: cooldown}, r.Status().Update(ctx, EvictionAutoScaler)
+		return r.handleBlockedDrain(ctx, EvictionAutoScaler, pdb, target, surgeApplier, maxSurgeTarget)
```
skuchipudi295, line 276: please dont do refactor and add logic in same pr

paulgmiller:
>rejecting maxUnavailable would exclude the common shape these workloads actually use,

Why do the use them though what are they trying to accomplish? With
out knowing that how do we know even a temporary conversion is safe?
I seem to have lost my comment I left on your pr description but it
claims maxunavailable is a common sfe deployment requirement. Whose
requirement? Why are they requiring it? what are they trying to
accomplish. Why does a minAvailable N% not work for them? Are these
intenral or external customers? Maybe aks could pull actual metrics on
how many customers actually use pdbs like this?

PR descripton also mentions "or an absolute minAvailable" isn't that not a problem unless their
minavailale is ABOVE the replcas? Is that an error? should we enourage
those people to use autogenerated pdbs instead

>> mis-protect the workload outside of drains.  

I don't think we know this because we don't know why customers 
are doing it. I think having manually created pdbs dynamically change is actually more confusing. If you want them dynanmic let enviction autoscaler auto generate them?


>> also percentage `minAvailable` 

You didn't talke about this in PR description.  Need to capture the motivations in the PR descripton or an issue for posterity.

skuchipudi295:
The per-namespace opt-in annotation (`eviction-autoscaler.azure.com/pdb-floor-mutation`) feels unnecessary given the existing gates.

By the time PDB floor mutation runs, the namespace has already opted into eviction-autoscaler (`eviction-autoscaler.azure.com/enable: "true"`) — consenting to replica surges, scale-downs, PDB creation, and pod evictions. The fleet-wide `ENABLE_PDB_FLOOR_MUTATION` env var is already a strong operator-level gate (startup-validated, requires controller restart to change).

The temporary, self-reverting PDB pin is arguably less invasive than what the controller already does without a second namespace annotation (e.g. `PDB_CREATE` creates permanent PDBs with only the existing namespace enable). Adding a second namespace annotation increases adoption friction and debugging surface ("why is pinning not working?") without a proportional safety gain.

If namespace-level granularity is needed, the existing enable annotation already partitions namespaces. If PDB-author-level consent is the goal, a per-PDB opt-out annotation would be more useful — it lets the actual PDB author (who may differ from the namespace owner) control their own object, rather than requiring namespace owners to consent on their behalf.

Suggestion: drop the namespace annotation gate and rely on `ENABLE_PDB_FLOOR_MUTATION` as the sole opt-in, or replace the namespace annotation with a per-PDB opt-out for cases where a specific PDB should be excluded.

skuchipudi295:
The stale mutation backstop may never actually fire in practice.

Every restore path is already covered: drain-complete (scale-down + idle), CR deletion (finalizer), and feature-disabled-on-restart. The backstop runs inside the CR reconciler, so it only fires when the CR exists and is being reconciled — but if reconciles are running, the normal drain-complete paths should already be converging.

The one scenario it seems designed for — an orphaned PDB with no CR — is the one it can't catch, since there's no CR to trigger the reconcile. Catching that would require a separate controller watching PDBs for stale `mutated-at` annotations.

Not harmful to keep, but worth noting it's defense-in-depth against a narrow code-bug scenario (reconcile running but no branch converging) rather than the orphaned-mutation case.

skuchipudi295:
Suggest moving `defendPinnedFloor` to a separate follow-up PR.

The core pin-and-restore lifecycle (pin before surge, restore on scale-down/deletion) stands on its own. The defend logic adds a second concern — what happens when an external actor overwrites the PDB mid-drain — and the right behavior there is debatable:

- For human edits (e.g. emergency PDB freeze during an incident), re-pinning overrides intentional partner action. The partner should have veto power during a drain.
- For GitOps/automation clobber (ArgoCD, Flux reverting PDB to declared state), re-pinning is essential or the feature breaks in GitOps-managed namespaces.
- The re-snapshot-on-defend also masks a stale-snapshot risk on the restore path that deserves its own design attention.

These are distinct trade-offs worth separate review. Splitting it out keeps this PR focused on the core lifecycle and lets the defend behavior be evaluated (and possibly gated or nuanced — e.g. only re-pin when the edit looks automated) on its own.

skuchipudi295:
On the restore path in `revertPDBFloor`: if `Update(PDB)` returns a conflict, just re-read the PDB, strip our annotations (`original-pdb-spec`, `mutated-at`, `pinned-floor`), and proceed to remove the finalizer + clear status. Don't attempt to restore the snapshot spec — the conflict means the customer wrote to the PDB, and that's their intent. Clean up and move on.

skuchipudi295:
Minor issue in `ensurePDBFloor`: the finalizer is added *before* the `pdbCarriesFloor` short-circuit check. If the PDB already carries the floor (e.g. customer authored `minAvailable: 4` matching F), the short-circuit returns without writing any annotations or mutating the spec — but the finalizer is already on the CR.

This means the CR carries a restore finalizer when nothing was mutated. It converges (restore is a no-op, finalizer gets cleaned up eventually), but it's unnecessary work and a confusing signal.

Fix: move the finalizer add to after the `pdbCarriesFloor` check, right before `snapshotPDBSpec`. Only add the finalizer when a mutation is actually about to happen.

skuchipudi295:
In `restoreFloorIfDrainHandled`, the cleanup check only looks at `isMutated(pdb)` and `PinnedPDBFloor`. If both are nil/false but the finalizer is still on the CR (possible if Status().Update failed on the initial surge write and the subsequent finalizer removal also hit a conflict), the idle path becomes a no-op and the finalizer stays indefinitely.

The finalizer should be included in the cleanup check:

```go
needsCleanup := isMutated(pdb) || 
                eas.Status.PinnedPDBFloor != nil ||
                controllerutil.ContainsFinalizer(eas, PDBFloorFinalizer)
if !needsCleanup {
    return nil
}
```

The finalizer should be removed alongside the annotations — if there's nothing to restore, there's nothing to finalize.

NamanBhat:
> On the restore path in `revertPDBFloor`: if `Update(PDB)` returns a conflict, just re-read the PDB, strip our annotations... and proceed... Don't attempt to restore the snapshot spec — the conflict means the customer wrote to the PDB.

Thanks Sushma. I looked at this closely and I think the current flow already converges safely on a conflict, so I'd propose keeping it as-is:

- The restore branch is gated by `pdbCarriesFloor(pdb, floor)` — we only push the snapshot when the PDB **still carries our exact pin**. If the partner changed the floor, we fall through to the `isMutated` branch, which does exactly what you describe: drop our tracking annotations and keep their live spec.
- On a 409 we return and requeue — the next reconcile re-reads the PDB fresh and re-runs that same `pdbCarriesFloor` check, so a concurrent partner/GitOps write is reclassified on the retry rather than blindly re-restored.
- The restore itself is scoped to only the two fields we manage (`minAvailable`/`maxUnavailable`); it never rewrites `Selector` or `UnhealthyPodEvictionPolicy`, so a concurrent edit to any other field survives untouched. Re-applying the original floor is also idempotent, so a forced retry is a no-op beyond the revert we intended.

Net: a conflict here self-heals without clobbering partner intent. The one thing your suggestion improves is not surfacing the benign race as a reconcile error — happy to swallow the 409 (log at `V(1)` + requeue) if you'd prefer that cosmetic cleanup. Also note there's no finalizer in this PR, so that part of the cleanup doesn't apply here.

PR #151, opened 2026-07-20
Controller Test: DS-blocked surge behavior

danekeahi:
#### Impact
Adds a controller-level unit test proving the eviction-autoscaler reconciler handles a Deployment Safeguards (DS) rejection of its surge write without panicking or crashing, across all three surge paths (Deployment, HPA, KEDA). Question to answer: if the controller surges in an AKS-owned namespace but a blocker (DS) rejects the write, is our code good enough to handle the error/panic?

#### How it simulates DS
Real DS on Automatic clusters is a ValidatingAdmissionPolicy + SAR restrictions that flat-out rejects ANY write to a protected/AKS-owned namespace, regardless of resource type. The test uses a controller-runtime interceptor, forbidWritesInNamespace, that returns Forbidden on every write to the protected namespace — faithful to that behavior, and it makes the three surge paths genuinely distinct write points.

#### What it does
- Drives the real reconciler through envtest (real API server, no fake appliers).
- Three specs: KEDA ScaledObject path, standalone HPA path, and Deployment path.
- Each cordons/triggers a surge, asserts the surge write is blocked (Forbidden), and asserts the reconciler NotTo(Panic()).
- Shared assertBlockedSurge helper keeps the three specs consistent.

#### Testing
- gofmt + go vet ./internal/controller/ — clean.
- go test ./internal/controller/ :

  Ran 87 of 87 Specs in 12.246 seconds
  SUCCESS! -- 87 Passed | 0 Failed | 0 Pending | 0 Skipped
  --- PASS: TestControllers (12.25s)

- Anti-vacuous proofs: flipping the interceptor to allow the write makes all three specs fail, and injecting a panic on the surge path makes all three fail — confirming the specs actually exercise the block + panic-safety.

#### Notes
- The e2e counterpart (real controller vs. a real admission webhook on kind) is a separate PR #152.

### DIFF
--- a/internal/controller/surge_ds_block_test.go
+++ b/internal/controller/surge_ds_block_test.go
@@ -28,12 +28,15 @@ func forbiddenOnUpdateClient(objs ...client.Object) client.Client {
 	Expect(kedav1alpha1.AddToScheme(scheme)).To(Succeed())
 	base := fake.NewClientBuilder().WithScheme(scheme).WithObjects(objs...).Build()
 	return interceptor.NewClient(base, interceptor.Funcs{
-		Update: func(_ context.Context, _ client.WithWatch, obj client.Object, _ ...client.UpdateOption) error {
-			return apierrors.NewForbidden(
-				schema.GroupResource{Resource: "deployments"},
-				obj.GetName(),
-				nil,
-			)
+		Update: func(ctx context.Context, c client.WithWatch, obj client.Object, opts ...client.UpdateOption) error {
+			if _, ok := obj.(*appsv1.Deployment); ok {
+				return apierrors.NewForbidden(
+					schema.GroupResource{Group: "apps", Resource: "deployments"},
+					obj.GetName(),
+					nil,
+				)
+			}
+			return c.Update(ctx, obj, opts...)
 		},
 	})
 }
--- a/internal/controller/surge_ds_block_test.go
+++ b/internal/controller/surge_ds_block_test.go
@@ -1,0 +1,93 @@
+package controllers
+
+import (
+	"context"
+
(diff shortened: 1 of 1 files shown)


internal/controller/surge_ds_block_test.go
```diff
@@ -0,0 +1,100 @@
...
+			}
+			return c.Update(ctx, obj, opts...)
+		},
+	})
+}
+
+// When Deployment Safeguards blocks the surge write, each surge strategy must return the
+// error (so the reconcile loop requeues) and must not panic. A panic here would fail the
+// spec because Ginkgo treats panics as failures.
+var _ = Describe("Surge appliers when Deployment Safeguards blocks the write", func() {
```
skuchipudi295: I also want to see how app e2e handles this.. testing surge applier might not be enough

skuchipudi295 (changes requested):

internal/controller/surge_ds_block_test.go
skuchipudi295: also need to test for panics.. for the same path

danekeahi: Created a new PR #152 for e2e testing, and I addressed panics. Each of the three specs now asserts NotTo(Panic()) on the reconcile call for the blocked surge path. #152 backs that at the app level by asserting the controller-manager pod never crashes or restarts (restartCount stays at baseline) while DS blocks the write. Verified both ways: unit suite passes 87/87 and e2e suite 6/6.

skuchipudi295:
### Code review

Reviewed at commit 3840528cecc4967a5c71225ea2c71e0de2e403b1.

skuchipudi295 (approved):

PR #153, opened 2026-07-25
Add panic metric with namespace and target labels

danekeahi:
#### Impact
Adds a panic counter so we can alert when the eviction autoscaler panics while reconciling. Today a reconcile panic is invisible in production: controller-runtime recovers it, so the pod never crashes or restarts and nothing pages, while that controller's work silently stops for the affected object.

#### Why not the built-in metric
controller-runtime emits `controller_runtime_reconcile_panics_total`, but it doesn't work here. Its only label is `controller`, so it can't say which workload panicked. More importantly, I checked the AKSCCPMetrics Kusto database: there is no `ControllerRuntimeReconcilePanicsTotal` table, and eviction-autoscaler doesn't appear in the controller-runtime tables at all (only gpu_provisioner and kaito_workspace do). Our own `eviction_autoscaler_*` metrics are already ingested as 9 tables, so emitting our own panic metric uses an ingestion path that already works.

#### What changed
- `eviction_autoscaler_panics_total` labeled by `namespace`, `target_name`, and `controller`.
- A small `recordPanic` helper that records the panic and re-panics, so controller-runtime's recovery still applies and runtime behavior is unchanged.
- One deferred call in each of the five reconcilers. The eviction autoscaler reconciler passes the spec target so panics on the surge path name the workload being surged.

#### Testing
- gofmt, go vet, and go build clean; full unit suite passes.
- Two new tests: one asserts the counter increments with the correct namespace/target and that the panic is re-raised; one asserts nothing is recorded on a normal return.
- Anti-vacuous proofs: recording the wrong namespace fails the test, and swallowing the panic fails the test. Restored, both pass.
- No dependency changes (go.mod and go.sum untouched).

#### Follow-up
A companion alert in aks-rp-alerts queries `EvictionAutoscalerPanicsTotal1mIncrease`. It can only fire once this ships, since the table won't exist until the metric is emitted in production.

### DIFF
--- a/internal/controller/evictionautoscaler_controller.go
+++ b/internal/controller/evictionautoscaler_controller.go
@@ -107,12 +107,12 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
 		return ctrl.Result{}, err
 	}
 
-	panicTarget = EvictionAutoScaler.Spec.TargetName
 	if EvictionAutoScaler.Spec.TargetName == "" {
 		degraded(&EvictionAutoScaler.Status.Conditions, "EmptyTarget", "no specified target")
 		logger.Error(err, "no specified target name", "targetname", EvictionAutoScaler.Spec.TargetName)
 		return ctrl.Result{}, r.Status().Update(ctx, EvictionAutoScaler)
 	}
+	panicTarget = EvictionAutoScaler.Spec.TargetName
 
 	// StatefulSets are intentionally skipped — their ordered pod management
 	// semantics conflict with the eviction surge strategy.
--- a/internal/controller/evictionautoscaler_controller.go
+++ b/internal/controller/evictionautoscaler_controller.go
@@ -51,7 +51,24 @@ const cooldown = 1 * time.Minute
 // +kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch
 // +kubebuilder:rbac:groups=autoscaling,resources=horizontalpodautoscalers,verbs=get;list;watch;update
 
+// recordPanic counts a recovered reconcile panic against the namespace and target that caused
+// it, then re-panics so controller-runtime still handles it as before.
+func recordPanic(controller, namespace string, target *string) {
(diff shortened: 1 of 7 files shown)


internal/controller/evictionautoscaler_controller.go
```diff
@@ -90,6 +107,7 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
 		return ctrl.Result{}, err
 	}
 
+	panicTarget = EvictionAutoScaler.Spec.TargetName
```
skuchipudi295: it should be after line 116

internal/controller/node_reconciler.go
```diff
@@ -36,6 +36,11 @@ const NodeNameIndex = "spec.nodeName"
 
 // Reconcile is the main loop of the controller. It will look for unschedulded nodes and for every pod on the node
 func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	namespace := req.Namespace
+	if namespace == "" {
```
skuchipudi295, line 40: why set it _cluster? if namespace == "" we have a bigger problem here

skuchipudi295:
### Code review

internal/controller/node_reconciler.go
skuchipudi295: oo this is for node reconciler. nvr mind

skuchipudi295 (changes requested):

internal/controller/evictionautoscaler_controller.go
danekeahi: Moved, now an empty target keeps req.Name for the panic label instead of being overwritten with an empty string.

skuchipudi295:
### Code review

Reviewed at commit d7fec0e88ce20f231aff2c9d7029d11986801459.

skuchipudi295 (approved):

skuchipudi295 (approved):

PR #158, opened 2026-08-04
fix: detect node drains driven by Karpenter/cluster-autoscaler taints, fix inverted cordon predicate

### LINKED ISSUE #157: Node drains driven by Karpenter (and cluster-autoscaler) are never detected: surge never fires, controller-created PDBs block evictions forever
## Summary

The node reconciler's only drain signal is the classic cordon (`node.spec.unschedulable`). Karpenter v1 **never cordons**: both graceful disruption and termination apply the `karpenter.sh/disrupted:NoSchedule` taint and start evicting, while `spec.unschedulable` stays `false` (see Karpenter's disruption docs / `pkg/controllers/node/termination`). Cluster-autoscaler's scale-down behaves the same way with its `ToBeDeletedByClusterAutoscaler` taint.

On such clusters the controller deadlocks itself:

1. With `PDB_CREATE=true` it creates PDBs with `minAvailable = replicas` — for 1-replica deployments that is `allowedDisruptions = 0`, so every eviction is refused by design until the surge lifts it.
2. The surge (1→2) is triggered from the node reconciler on drain detection — which never fires, because the trigger is cordon-only.
3. Result: the node's eviction attempts fail forever. Karpenter retries indefinitely (continuous `FailedDraining` events), nodes stay in deleting state for months, clusters pay double capacity during rotations, and AMI security patching never completes.

Observed in production (EKS, Karpenter 1.10→1.12): every `EvictionAutoScaler` CR had an empty `spec.lastEviction` since creation (~5 months), nodes stuck deleting up to 126 days, pods only ever moved when a deploy's rollout happened to bypass the eviction API.

## Second, related bug: the node update predicate is inverted

`internal/controller/node_reconciler.go`:

```go
UpdateFunc: func(ue event.UpdateEvent) bool {
    oldNode := ue.ObjectOld.(*corev1.Node)
    newNode := ue.ObjectNew.(*corev1.Node)
    return oldNode.Spec.Unschedulable == newNode.Spec.Unschedulable
}
```

`==` returns `true` when the field did **not** change — i.e. the filter drops genuine cordon transitions and instead reconciles on every unrelated node update (heartbeats included). Cordons are currently only caught *laggily* via those unrelated updates.

## Proposed fix

Treat a known draining taint as equivalent to a cordon, and fix the predicate to trigger on drain-signal *transitions* (cordon flip, draining taint added/removed):

```go
var drainingTaintKeys = []string{
    "karpenter.sh/disrupted",            // Karpenter v1 disruption/termination
    "ToBeDeletedByClusterAutoscaler",    // cluster-autoscaler scale-down
}
```

Matching must be by specific key + `NoSchedule` effect — a generic "any NoSchedule taint" check would misfire on permanent dedicated-node taints (a standard pattern).

We are running this patch in production (fork: https://github.com/kylan11/eviction-autoscaler/tree/dbridge-1.0.0, based on 1.0.0). PR with the fix + tests: incoming from `fix/karpenter-taint-trigger` (based on current main) once we finish validating it on our staging clusters.

kylan11:
Fixes #157

## Problem

The node reconciler's only drain signal is `node.spec.unschedulable`. Karpenter v1 never cordons — disruption and termination apply the `karpenter.sh/disrupted:NoSchedule` taint while `spec.unschedulable` stays false. Cluster-autoscaler's scale-down uses `ToBeDeletedByClusterAutoscaler` the same way. On those clusters the drain is never detected, the surge never fires, and the controller's own `minAvailable = replicas` PDBs block every eviction indefinitely (months-stuck deleting nodes, continuous `FailedDraining`, double capacity during rotations — details and production evidence in the issue).

Additionally, the node update predicate was inverted: `return oldNode.Spec.Unschedulable == newNode.Spec.Unschedulable` returns true when the field did **not** change, so genuine cordon transitions were discarded and the controller reconciled on every unrelated node update instead.

## Changes

- `hasDrainingTaint` / `nodeIsDraining`: a node is considered draining if cordoned **or** carrying a known draining taint (`karpenter.sh/disrupted`, `ToBeDeletedByClusterAutoscaler`). Matching is by specific key + `NoSchedule` effect: a generic any-NoSchedule check would misfire on permanent dedicated-node taints.
- `Reconcile` gates on `nodeIsDraining` (and no longer returns a stale `err` on the early-return path).
- The update predicate is extracted to `nodeDrainSignalChanged` and now triggers on drain-signal *transitions* (cordon flip, draining taint added/removed) with the comparison direction fixed. Create/Delete events keep the default so already-draining nodes reconcile on informer sync.

## Testing

- New envtest case: node tainted `karpenter.sh/disrupted` with `unschedulable=false` → pod gets the `DisruptionTarget` condition and the `EvictionAutoScaler`'s `spec.lastEviction` is populated (fails on main).
- New case: node with only a permanent dedicated-node taint (`env=system:NoSchedule`) is ignored.
- Predicate table test: cordon flips and draining-taint transitions trigger; heartbeats/unrelated updates and non-Node objects don't.
- Validated on a live EKS cluster (Karpenter 1.12): on `kubectl delete nodeclaim`, `spec.lastEviction` was populated the same second the taint landed, the eviction proceeded and the node terminated in ~60s; with the previous behavior the same flow deadlocked indefinitely.

### DIFF
--- a/config/crd/bases/eviction-autoscaler.azure.com_evictionautoscalers.yaml
+++ b/config/crd/bases/eviction-autoscaler.azure.com_evictionautoscalers.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.20.1
+    controller-gen.kubebuilder.io/version: v0.21.0
   name: evictionautoscalers.eviction-autoscaler.azure.com
 spec:
   group: eviction-autoscaler.azure.com
--- a/internal/controller/evictionautoscaler_controller.go
+++ b/internal/controller/evictionautoscaler_controller.go
@@ -194,7 +194,11 @@ func (r *EvictionAutoScalerReconciler) Reconcile(ctx context.Context, req ctrl.R
 			EvictionAutoScaler.Status.MinReplicas = minReplicas
 		}
 		ready(&EvictionAutoScaler.Status.Conditions, "TargetSpecChange", fmt.Sprintf("resetting min replicas to %d", EvictionAutoScaler.Status.MinReplicas))
-		return ctrl.Result{}, r.Status().Update(ctx, EvictionAutoScaler) //should we go rety in case there is also an eviction or just wait till the next eviction
+		// Requeue explicitly: status updates don't bump metadata.generation, so the
+		// generation-only event filter won't re-trigger us. Returning without a
+		// requeue here strands an active surge at the scaled-out state until the
+		// next spec change (i.e. the next eviction), which may be days away.
+		return ctrl.Result{RequeueAfter: cooldown}, r.Status().Update(ctx, EvictionAutoScaler)
 	}
 
 	// Log current state before checks
--- a/internal/controller/node_reconciler.go
+++ b/internal/controller/node_reconciler.go
@@ -31,6 +31,38 @@ type NodeReconciler struct {
 
 const NodeNameIndex = "spec.nodeName"
 
+// drainingTaintKeys are taints applied by node-lifecycle controllers that drain
+// nodes without cordoning them (spec.unschedulable stays false):
+//   - karpenter.sh/disrupted: Karpenter v1 disruption and termination
+//   - ToBeDeletedByClusterAutoscaler: cluster-autoscaler scale-down
+var drainingTaintKeys = []string{
+	"karpenter.sh/disrupted",
+	"ToBeDeletedByClusterAutoscaler",
+}
+
+// hasDrainingTaint reports whether the node carries a NoSchedule taint from a
+// known draining controller. Matching specific keys (not any NoSchedule taint)
+// matters: dedicated-node taints are permanent and do not signal a drain.
+func hasDrainingTaint(node *corev1.Node) bool {
+	for _, taint := range node.Spec.Taints {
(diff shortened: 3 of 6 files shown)


kylan11:
Pushed a second commit: validation on a live Karpenter cluster showed `countPodsOnCordoned` (surge sizing added in #124/1.0.0) still keys on `Spec.Unschedulable` alone, so on taint-drained nodes `displaced` stayed 0 and the scale-up never fired even though `spec.lastEviction` was recorded. The commit extends the same drain signal (`nodeIsDraining`) to that helper, with unit tests for the tainted and dedicated-node cases.

kylan11:
Third commit pushed, found during production validation of the first two: the generation-sync branch in `evictionautoscaler_controller.go` returns without a requeue, and since status updates don't bump `metadata.generation`, the generation-only event filter never re-triggers reconciliation. Net effect: after a surge (which bumps the target's generation), the revert branch is unreachable until the *next* eviction — the deployment stays scaled out indefinitely (we observed 2 replicas stranded 20+ min after the drained node terminated, with the PDB already allowing disruptions). Fixed by returning `RequeueAfter: cooldown` from that branch. Note this is hard to catch in the existing tests because they invoke `Reconcile` directly and don't exercise the requeue chain.

skuchipudi295 (changes requested):
please create an ICM and run feature requests by RP team before making changes or submitting pull requests.

PR #159, opened 2026-08-04
feat(helm): add nodeSelector, affinity, and tolerations for the contr…

skuchipudi295:
This pull request adds support for advanced Kubernetes pod scheduling options for the controller pod, allowing users to specify `nodeSelector`, `affinity`, and `tolerations` via Helm values or Azure CLI configuration. These changes make it possible to control which nodes the controller pod runs on, including targeting specific node pools or tainted nodes. Documentation has been updated with usage examples.

**Controller pod scheduling enhancements:**

* Added `nodeSelector`, `affinity`, and `tolerations` fields to the Helm chart values (`values.yaml`) to allow users to control pod placement and node selection for the controller.
* Modified the controller `Deployment` template (`deployment.yaml`) to use the values of `nodeSelector`, `affinity`, and `tolerations` if provided.

**Documentation updates:**

* Updated the `README.md` with explanations and examples for configuring controller pod scheduling, including both Azure CLI and Helm usage.…oller deployment

Expose pod scheduling configuration for the eviction-autoscaler controller so it can be pinned to specific nodes (node affinity by label) and scheduled onto tainted nodes. Defaults are no-ops, so existing installs are unaffected. Documented in values.yaml and the README config section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,36 @@ az k8s-extension create \
 - `controllerConfig.pdb.create=true` - Automatically creates PDBs for deployments (default: false)
 - `controllerConfig.namespaces.enabledByDefault=true` - Enables all namespaces (default: false, opt-in mode)
 - `controllerConfig.namespaces.actionedNamespaces` - List of namespaces to enable when using opt-in mode (default: [kube-system])
+- `nodeSelector` - Pins the controller pod to nodes matching these labels (default: `{}`, no constraint)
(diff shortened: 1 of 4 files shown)


NamanBhat (approved):
LGTM - clean, non-breaking chart change adding the standard nodeSelector / affinity / tolerations plumbing on the controller pod. Correct Helm usage, inert defaults ({} / []) so existing installs are unaffected, and the README default note matches values.yaml. Gives operators the standard knobs to pin the controller to specific nodes (e.g. a dedicated nodepool via a node label plus a matching toleration). Thanks!

PR #160, opened 2026-08-05
feat: add global controller kill switch and AKS-owned namespace opt-out

NamanBhat:
## What

Adds two operator controls for disabling / narrowing what the controller acts on, both defaulting to preserve current behavior.

### 1. Global controller kill switch — `CONTROLLER_ENABLED` (`controllerConfig.enabled`, default `true`)
When `false`, **no reconcilers are registered** — the controller still runs and serves health/metrics, but takes no action on any namespace or PDB. This lets operators fully disable eviction-autoscaler **in place** (via config) without uninstalling the extension. Today the only way to stop the controller is to uninstall.

### 2. AKS-owned namespace opt-out — `MANAGE_AKS_OWNED_NAMESPACES` (`controllerConfig.namespaces.manageAKSOwnedNamespaces`, default `true`)
Today AKS-owned namespaces (`kube-system`, `flux-system`, etc.) are **unconditionally managed** — the namespace filter short-circuits `IsAKSOwnedNamespace()` before consulting `enabledByDefault` / `actionedNamespaces` / the annotation, so they cannot be excluded. When this is set to `false`, those namespaces follow the **normal enable rules** like any other namespace, so operators who want the controller to act only on an explicitly opted-in set can exclude the system namespaces entirely.

## Why

Fleet operators asked for (a) a single in-place disable that doesn't require an uninstall, and (b) a way to keep the controller from acting on system namespaces they don't want it touching. Both are additive, opt-in, and default to today's behavior.

## Changes

- `cmd/main.go` — parse `CONTROLLER_ENABLED` and `MANAGE_AKS_OWNED_NAMESPACES` (strict bool, fail-fast on bad value); gate reconciler registration on the kill switch; wire the filter option.
- `internal/namespacefilter/nsfilter.go` — backward-compatible `WithManageAKSOwnedNamespaces` functional option (default keeps unconditional management); gate the AKS-owned short-circuit on it.
- `internal/namespacefilter/nsfilter_test.go` — unit tests: default still manages `kube-system`; excluded when the option is off in opt-in mode; annotation still re-enables when off.
- `helm/eviction-autoscaler/{values.yaml,templates/deployment.yaml}` — new settings + env wiring.
- `README.md` — documented both.

## Compatibility

Both flags default to `true`, so existing installs are unaffected. `New(...)`'s existing call sites are unchanged (the new behavior is a variadic option).

## Testing

- `go vet ./cmd/... ./internal/...` — OK
- `go test ./internal/namespacefilter/...` — passing (incl. 3 new tests)
- `helm` values validated

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,9 +151,11 @@ az k8s-extension create \
 
 **Configuration Options:**
 
+- `controllerConfig.enabled=false` - Global kill switch; when `false`, no reconcilers run and the controller takes no action (still serves health/metrics) (default: true)
 - `controllerConfig.pdb.create=true` - Automatically creates PDBs for deployments (default: false)
 - `controllerConfig.namespaces.enabledByDefault=true` - Enables all namespaces (default: false, opt-in mode)
 - `controllerConfig.namespaces.actionedNamespaces` - List of namespaces to enable when using opt-in mode (default: [kube-system])
+- `controllerConfig.namespaces.manageAKSOwnedNamespaces=false` - When `false`, AKS-owned namespaces (kube-system, flux-system, etc.) follow the normal enable rules instead of being unconditionally managed, so they can be excluded (default: true)
 
 **Common Configuration Combinations:**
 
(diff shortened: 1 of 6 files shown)


NamanBhat:
Split into two focused PRs for easier review: #162 (global controller kill switch) and #161 (AKS-owned namespace opt-out). Closing this combined PR in favor of those.

PR #161, opened 2026-08-05
feat: add opt-out for managing AKS-owned namespaces

NamanBhat:
## What

Adds an opt-out for unconditionally managing AKS-owned namespaces.

`MANAGE_AKS_OWNED_NAMESPACES` (`controllerConfig.namespaces.manageAKSOwnedNamespaces`, default `true`).

Today AKS-owned namespaces (`kube-system`, `flux-system`, etc.) are **unconditionally managed** — the namespace filter short-circuits `IsAKSOwnedNamespace()` before consulting `enabledByDefault` / `actionedNamespaces` / the annotation, so they cannot be excluded. When this is set to `false`, those namespaces follow the **normal enable rules** like any other namespace, so operators who want the controller to act only on an explicitly opted-in set can exclude the system namespaces entirely.

## Why

Fleet operators need the controller to act only on an explicitly opted-in namespace set; today there is no way to keep it off the AKS-owned/system namespaces. This is additive, opt-in, and defaults to today's behavior.

## Changes

- `internal/namespacefilter/nsfilter.go` — backward-compatible `WithManageAKSOwnedNamespaces` functional option (default keeps unconditional management); gate the AKS-owned short-circuit on it.
- `internal/namespacefilter/nsfilter_test.go` — unit tests: default still manages `kube-system`; excluded when the option is off in opt-in mode; annotation still re-enables when off.
- `cmd/main.go` — parse `MANAGE_AKS_OWNED_NAMESPACES` (strict bool, fail-fast); wire the filter option.
- `helm/eviction-autoscaler/{values.yaml,templates/deployment.yaml}` — new setting + env wiring.
- `README.md` — documented.

## Compatibility

Defaults to `true`, so existing installs are unaffected. Existing `New(...)` call sites are unchanged (the new behavior is a variadic option).

## Testing

- `go test ./internal/namespacefilter/...` — passing (incl. 3 new tests)

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ az k8s-extension create \
 - `controllerConfig.pdb.create=true` - Automatically creates PDBs for deployments (default: false)
 - `controllerConfig.namespaces.enabledByDefault=true` - Enables all namespaces (default: false, opt-in mode)
 - `controllerConfig.namespaces.actionedNamespaces` - List of namespaces to enable when using opt-in mode (default: [kube-system])
+- `controllerConfig.namespaces.manageAKSOwnedNamespaces=false` - When `false`, AKS-owned namespaces (kube-system, flux-system, etc.) follow the normal enable rules instead of being unconditionally managed, so they can be excluded (default: true)
 
 **Common Configuration Combinations:**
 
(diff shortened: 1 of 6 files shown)


skuchipudi295:
as discussed in thread internally, cant be supported as it goes against what AKS wants

PR #162, opened 2026-08-05
feat: add global controller kill switch (CONTROLLER_ENABLED)

NamanBhat:
## What

Adds a global controller kill switch: `CONTROLLER_ENABLED` (`controllerConfig.enabled`, default `true`).

When `false`, **no reconcilers are registered** — the controller still runs and serves health/metrics, but takes no action on any namespace or PDB. This lets operators fully disable eviction-autoscaler **in place** (via config) without uninstalling the extension. Today the only way to stop the controller is to uninstall.

## Why

Fleet operators need a fast, in-place way to disable the controller during an incident, without an uninstall-per-cluster. Additive, opt-in, and defaults to today's behavior.

## Changes

- `cmd/main.go` — parse `CONTROLLER_ENABLED` (strict bool, fail-fast on bad value); gate reconciler registration on it (when disabled, health/metrics still serve).
- `helm/eviction-autoscaler/{values.yaml,templates/deployment.yaml}` — new setting + env wiring.
- `README.md` — documented.

## Compatibility

Defaults to `true`, so existing installs are unaffected.

## Testing

## Note

Split out from #160 into two focused PRs (this one = kill switch; the AKS-owned namespace opt-out is a separate PR). #160 will be closed in favor of the two.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ az k8s-extension create \
 
 **Configuration Options:**
 
+- `controllerConfig.enabled=false` - Global kill switch; when `false`, no reconcilers run and the controller takes no action (still serves health/metrics) (default: true)
 - `controllerConfig.pdb.create=true` - Automatically creates PDBs for deployments (default: false)
 - `controllerConfig.namespaces.enabledByDefault=true` - Enables all namespaces (default: false, opt-in mode)
 - `controllerConfig.namespaces.actionedNamespaces` - List of namespaces to enable when using opt-in mode (default: [kube-system])
@@ -246,6 +247,7 @@ Eviction autoscaler provides flexible namespace-level control with two operation
 
 #### Environment Variables
 
+- **`CONTROLLER_ENABLED`**: Global kill switch for the controller (default: `true`). When `false`, no reconcilers are registered — the controller runs (serving health and metrics) but takes no action on any namespace or PDB, so the feature can be fully disabled in place without uninstalling the extension.
 - **`ENABLED_BY_DEFAULT`**: Controls the operational mode (default: `false`)
   - `false`: Namespaces disabled by default - only specified namespaces enabled
   - `true`: Namespaces enabled by default - all namespaces enabled unless disabled
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -209,58 +209,78 @@ func main() {
 	}
 	setupLog.Info("Zero-maxSurge override configuration", "zeroSurgeOverride", zeroSurgeOverride)
 
-	evictionAutoScalerReconciler := &controllers.EvictionAutoScalerReconciler{
-		Client:            mgr.GetClient(),
-		Scheme:            mgr.GetScheme(),
-		Filter:            nsfilter,
-		ZeroSurgeOverride: zeroSurgeOverride,
+	// Parse CONTROLLER_ENABLED environment variable (defaults to true). This is a
+	// global kill switch: when false, no reconcilers are registered, so the
+	// controller runs (serving health and metrics) but takes no action on any
+	// namespace or PDB — letting operators fully disable the feature in place via
+	// config, without uninstalling the extension.
+	controllerEnabled := true
+	if v := os.Getenv("CONTROLLER_ENABLED"); v != "" {
+		var err error
+		controllerEnabled, err = strconv.ParseBool(v)
+		if err != nil {
+			setupLog.Error(err, "Failed to parse CONTROLLER_ENABLED env variable")
+			os.Exit(1)
+		}
 	}
-	if err = evictionAutoScalerReconciler.SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "EvictionAutoScaler")
-		os.Exit(1)
-	}
-	setupLog.Info("EvictionAutoScalerReconciler setup completed")
+	setupLog.Info("Controller enable configuration", "controllerEnabled", controllerEnabled)
 
-	if pdbCreate {
-		if err = (&controllers.DeploymentToPDBReconciler{
+	if controllerEnabled {
+		evictionAutoScalerReconciler := &controllers.EvictionAutoScalerReconciler{
+			Client:            mgr.GetClient(),
+			Scheme:            mgr.GetScheme(),
+			Filter:            nsfilter,
+			ZeroSurgeOverride: zeroSurgeOverride,
+		}
+		if err = evictionAutoScalerReconciler.SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "EvictionAutoScaler")
+			os.Exit(1)
+		}
+		setupLog.Info("EvictionAutoScalerReconciler setup completed")
+
+		if pdbCreate {
+			if err = (&controllers.DeploymentToPDBReconciler{
+				Client: mgr.GetClient(),
+				Scheme: mgr.GetScheme(),
+				Filter: nsfilter,
+			}).SetupWithManager(mgr); err != nil {
+				setupLog.Error(err, "unable to create controller", "controller", "DeploymentToPDBReconciler")
+				os.Exit(1)
+			}
+			setupLog.Info("DeploymentToPDBReconciler setup completed")
+
+			// Watches both HPA and KEDA ScaledObject changes to keep PDB minAvailable
+			// in sync with the autoscaler's min replicas floor.
+			if err = (&controllers.AutoscalerToPDBReconciler{
+				Client: mgr.GetClient(),
+				Scheme: mgr.GetScheme(),
+				Filter: nsfilter,
+			}).SetupWithManager(mgr); err != nil {
+				setupLog.Error(err, "unable to create controller", "controller", "AutoscalerToPDBReconciler")
+				os.Exit(1)
+			}
+			setupLog.Info("AutoscalerToPDBReconciler setup completed")
+		}
+
+		if err = (&controllers.PDBToEvictionAutoScalerReconciler{
 			Client: mgr.GetClient(),
 			Scheme: mgr.GetScheme(),
 			Filter: nsfilter,
 		}).SetupWithManager(mgr); err != nil {
-			setupLog.Error(err, "unable to create controller", "controller", "DeploymentToPDBReconciler")
+			setupLog.Error(err, "unable to create controller", "controller", "PDBToEvictionAutoScalerReconciler")
 			os.Exit(1)
 		}
-		setupLog.Info("DeploymentToPDBReconciler setup completed")
+		setupLog.Info("PDBToEvictionAutoScalerReconciler  setup completed")
 
-		// Watches both HPA and KEDA ScaledObject changes to keep PDB minAvailable
-		// in sync with the autoscaler's min replicas floor.
-		if err = (&controllers.AutoscalerToPDBReconciler{
+		if err = (&controllers.NodeReconciler{
 			Client: mgr.GetClient(),
 			Scheme: mgr.GetScheme(),
-			Filter: nsfilter,
 		}).SetupWithManager(mgr); err != nil {
-			setupLog.Error(err, "unable to create controller", "controller", "AutoscalerToPDBReconciler")
+			setupLog.Error(err, "unable to create controller", "controller", "NodeReconciler")
 			os.Exit(1)
 		}
-		setupLog.Info("AutoscalerToPDBReconciler setup completed")
-	}
-
-	if err = (&controllers.PDBToEvictionAutoScalerReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
-		Filter: nsfilter,
-	}).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "PDBToEvictionAutoScalerReconciler")
-		os.Exit(1)
-	}
-	setupLog.Info("PDBToEvictionAutoScalerReconciler  setup completed")
-
-	if err = (&controllers.NodeReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
-	}).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "EvictionAutoScaler")
-		os.Exit(1)
+	} else {
+		setupLog.Info("Controller is disabled (CONTROLLER_ENABLED=false); no reconcilers registered — serving health and metrics only")
 	}
 	// +kubebuilder:scaffold:builder
 
--- a/helm/eviction-autoscaler/templates/deployment.yaml
+++ b/helm/eviction-autoscaler/templates/deployment.yaml
@@ -43,6 +43,8 @@ spec:
             value: {{ join "," .Values.controllerConfig.namespaces.actionedNamespaces | quote }}
           - name: ZERO_SURGE_OVERRIDE
             value: {{ .Values.controllerConfig.zeroSurgeOverride | quote }}
+          - name: CONTROLLER_ENABLED
+            value: {{ .Values.controllerConfig.enabled | quote }}
         command:
         - /manager
         args:
--- a/helm/eviction-autoscaler/values.yaml
+++ b/helm/eviction-autoscaler/values.yaml
@@ -19,6 +19,13 @@ resources:
 
 # Controller configuration - the main settings users need to customize
 controllerConfig:
+  # Global enable/disable for the entire controller (kill switch).
+  # - true (default): reconcilers run normally.
+  # - false: no reconcilers are registered, so the controller takes no action on any
+  #          namespace or PDB (it still serves health and metrics). Use this to fully
+  #          disable eviction-autoscaler in place, without uninstalling the extension.
+  enabled: true
+
   # Metrics configuration
   metrics:
     enabled: true


cmd/main.go
```diff
@@ -209,58 +209,78 @@ func main() {
...
+	// config, without uninstalling the extension.
+	controllerEnabled := true
+	if v := os.Getenv("CONTROLLER_ENABLED"); v != "" {
+		var err error
+		controllerEnabled, err = strconv.ParseBool(v)
+		if err != nil {
+			setupLog.Error(err, "Failed to parse CONTROLLER_ENABLED env variable")
+			os.Exit(1)
+		}
 	}
```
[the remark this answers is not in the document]
NamanBhat: This is intentional for consistency — the other boolean env flags in this controller (ENABLED_BY_DEFAULT, PDB_CREATE) also use strconv.ParseBool, so CONTROLLER_ENABLED follows the same convention and accepts the standard ParseBool values. Keeping it consistent rather than special-casing this one flag as strict-only.

cmd/main.go
```diff
@@ -209,58 +209,78 @@ func main() {
...
 			Client: mgr.GetClient(),
 			Scheme: mgr.GetScheme(),
 			Filter: nsfilter,
 		}).SetupWithManager(mgr); err != nil {
-			setupLog.Error(err, "unable to create controller", "controller", "DeploymentToPDBReconciler")
+			setupLog.Error(err, "unable to create controller", "controller", "PDBToEvictionAutoScalerReconciler")
 			os.Exit(1)
 		}
-		setupLog.Info("DeploymentToPDBReconciler setup completed")
+		setupLog.Info("PDBToEvictionAutoScalerReconciler  setup completed")
```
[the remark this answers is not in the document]
NamanBhat: Fixed — removed the double space ("Reconciler  setup" -> "Reconciler setup"). Pushed in 0b0fb86.

skuchipudi295:
### Code review

Overall approach is clean — wrapping the reconciler registration block is the right way to do a kill switch. Two concerns:

**1. No graceful wind-down on disable**

When `CONTROLLER_ENABLED=false`, anything in-flight stays frozen:
- Surged deployments stay surged (annotation-marked replicas never scale back)
- Controller-created PDBs remain but stop tracking replica changes
- Pinned PDB floors (if #170 merges) never restore

System namespaces are typically where EAS is most active. Flipping the switch mid-drain could leave nodes stuck. At minimum, document that operators should verify no active surges exist before disabling. Ideally, add a startup cleanup pass that restores surged deployments and pinned PDBs when `CONTROLLER_ENABLED=false`.

**2. "Serves metrics" claim is misleading**

All operational metrics (`BlockedEvictionCounter`, `ScalingOpportunityCounter`, `ActualScalingCounter`, `PDBCounter`, etc.) are incremented inside reconciler code paths. With no reconcilers registered, none of these fire — the `/metrics` endpoint only serves Go runtime stats and zeros.

We should still observe what the system *sees* even when disabled (e.g. surge opportunities, PDB-blocked drains, displaced pods on cordoned nodes). Consider keeping a lightweight read-only watcher that increments observability counters without taking action, so operators can see what the controller *would* be doing while it is disabled. Otherwise, update the docs to clarify that only health probes are meaningful when disabled, not metrics.

skuchipudi295:
**Correction to my previous comment on metrics:**

The kill switch should not skip reconciler registration entirely. Controllers should still be registered so that watches fire and reconciles run in **observe-only mode** — metrics like `BlockedEvictionCounter`, `ScalingOpportunityCounter`, `PDBCounter` etc. continue to increment based on what the controller sees. What gets suppressed is all *writes*: no PDB creation, no EAS creation, no surge application, no PDB mutation.

Current implementation (wrapping the entire registration block in `if controllerEnabled`) kills observability along with actuation. Instead, pass `controllerEnabled` down to the reconcilers and gate the write paths, not the registration.

skuchipudi295:
**Additional notes:**

1. **Important metrics are skipped with the current approach.** Counters like `BlockedEvictionCounter`, `ScalingOpportunityCounter`, `PDBCounter`, and `ActualScalingCounter` are all incremented inside reconcile loops. By not registering the controllers at all, you lose visibility into what the system is experiencing. The existing `!isEnabled` namespace filter pattern already solves this — reconcilers run, observe, increment metrics, then early-return before writes. The kill switch should follow the same pattern at a global level rather than skipping registration.

2. **The kill switch requires a pod restart to take effect.** `CONTROLLER_ENABLED` is read once at startup. If an operator flips the Helm value, the pod must be restarted (rollout restart) for it to take effect. This should be documented — operators might expect a config change to propagate live. Consider whether a runtime-checked flag (e.g. a ConfigMap watch or leader election lease annotation) would be more appropriate for a "kill switch" that needs to act fast in an incident.

skuchipudi295:
**On the design direction:**

Skipping controller registration entirely makes this functionally equivalent to uninstalling the extension — except you are still paying for a pod that does nothing. If the goal is "no controllers run at all," operators can just uninstall.

The value of a kill switch *over* uninstalling is:
- **Hot toggle without restart** (per Teams discussion) — flip a ConfigMap/CR field, reconcilers immediately stop writing but keep observing
- **Preserve state** — EAS CRs, PDBs, annotations stay in place; re-enable picks up where it left off
- **Observability while disabled** — still see blocked drains, surge opportunities in metrics without acting
- **No RBAC/webhook churn** — no extension reinstall/CRD reapply

Suggest: back the flag with a watched ConfigMap (cache read is cheap per-reconcile), reconcilers check it every loop, skip writes when false. Same pattern as the namespace filter but global and hot-reloadable.

NamanBhat:
Thanks Sushma — grouping the responses by the threads you raised.

**On "serves metrics" (fair — fixed the wording).** You're right that the operational counters (`BlockedEvictionCounter`, `ScalingOpportunityCounter`, `PDBCounter`, `ActualScalingCounter`) only advance inside reconcile loops. To be precise: the `/metrics` endpoint *does* stay up when disabled — the manager's metrics server runs regardless of reconcilers, and the collectors are registered at init — but the operational counters don't *advance* while disabled, so it exposes only static/runtime metrics. I've reworded the README/values/comment to say exactly that instead of "serves metrics."

**On observe-only vs skipping registration.** I made this a **true halt** (no reconcilers, no watches) deliberately. The switch's purpose is a fast in-place disable during an incident and a disabled-by-default resting state — in both you want the controller genuinely inert, not still watching/reconciling. Keeping watches + reconciles running purely to advance metrics is real overhead when a large fleet may sit disabled by default. "Observe what EAS *would* do while not acting" is really a **dry-run/observe mode** — a distinct feature from an on/off switch — which I'm happy to propose separately if there's demand. (Two notes: gating writes wouldn't help the wind-down point below, since reverting a surge is itself a write; and for per-namespace visibility while active, the existing `!isEnabled` filter already observes + early-returns before writes.)

**On graceful wind-down.** Agreed this deserves to be explicit. By design this is a **halt, not a teardown**: when disabled, an in-flight surge or mutated/pinned PDB is left as-is and resumes converging on re-enable. It's safe — a pinned floor never drops availability below baseline and drains fail-open — but it isn't unwound. I've documented that (verify no active surges before disabling if you need a clean state). A startup cleanup pass that restores surges/pins on disable is a reasonable enhancement, but it's a distinct **teardown** behavior (it's actuation, so it can't sit behind the same "do nothing" flag) — I'd track it as a follow-up.

**On hot-reload / "equivalent to uninstalling."** Flipping the flag is an AKS extension **config-settings patch** (ARM `Update` / Helm upgrade), **not** an uninstall+install — the extension, CRDs, RBAC and webhooks all stay; only the manager pod rolls to pick up the new value, and state (CRs, PDBs, annotations) is preserved so re-enable resumes from it. That's the opposite of uninstalling. A live ConfigMap-watched flag is feasible, but for a fail-open, level-triggered, off-data-path controller a normal pod roll to apply config is harmless, and env is consistent with the project's other switches (`ENABLED_BY_DEFAULT`, `PDB_CREATE`). A watch only saves a benign restart (same propagation latency) at the cost of a watch/cache path + the observe-only refactor — better as a follow-up if we ever need sub-second, zero-restart disable.

paulgmiller:
In PR description talk about why changing a config var is easier/faster than uninstalling? Won't the extension just  revert the deployment if you change it directly with outoging through extension settings?

cmd/main.go
```diff
@@ -209,58 +209,78 @@ func main() {
...
+			Filter:            nsfilter,
+			ZeroSurgeOverride: zeroSurgeOverride,
+		}
+		if err = evictionAutoScalerReconciler.SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "EvictionAutoScaler")
+			os.Exit(1)
+		}
+		setupLog.Info("EvictionAutoScalerReconciler setup completed")
+
+		if pdbCreate {
```
paulgmiller, line 252: indentation creep.  https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

paulgmiller: one way to solve this would be to always make a list/map  of controllers then for each call setup with manager in a loop. If controller !enabled skip that loop.

paulgmiller: Or have manager setup in seperaet function that is skipped.

README.md
```diff
@@ -246,6 +247,7 @@ Eviction autoscaler provides flexible namespace-level control with two operation
 
 #### Environment Variables
 
+- **`CONTROLLER_ENABLED`**: Global kill switch for the controller (default: `true`). When `false`, no reconcilers are registered — the controller runs (serving health and metrics) but takes no action on any namespace or PDB, so the feature can be fully disabled in place without uninstalling the extension.
```
paulgmiller: might want to be clear for all these vars we only read at starup time so if you try and control with configmap you'll also need a restart for them to take effect.

skuchipudi295:
1. Removed the README additions
[README.md](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (reverted to pre-PR state)
Why: You asked to drop the doc changes from this PR — the two entries (controllerConfig.enabled bullet + CONTROLLER_ENABLED env-var section) are gone; README now matches [main](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) exactly.

2. Corrected the "disabled → metrics" comment in [main.go](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
[main.go](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — the kill-switch else branch
Why: My first comment claimed the controller "emits none of its own metrics; only controller-runtime baseline remains." Cross-checking the code disproved that: [metrics.go:213](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) registers all collectors in [init()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (import-time), so the metric families stay exposed even when disabled, and the plain [NodeCordoningCounter](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) emits a 0 series. The comment now states this accurately (families exposed but not updated; *Vec produce no series, plain counter sits at 0).

3. Added a controller_enabled gauge (the substantive fix)
Files: [metrics.go](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (new [ControllerEnabled](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) gauge + registration) and [main.go](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (named [metrics](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) import + Set(1)/Set(0) on both kill-switch paths)
Why: You flagged a real observability gap — toggling the switch is a Helm upgrade → pod rollout, after which the business-metric series vanish (Vec metrics with no reconciles = no series), which is indistinguishable from a crash/scrape failure, and up still reads 1 because [/metrics](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) keeps serving. So metrics across the disable boundary "don't make sense."
The new eviction_autoscaler_controller_enabled is a plain 0/1 gauge set once at startup on both branches, giving a continuous series across disable/enable. That lets dashboards explain the metric cliff and lets alerts gate on controller_enabled == 1 (so an intentional disable doesn't page), cleanly separating "disabled by config" from "process down."

skuchipudi295:
need elaborate why

skuchipudi295:
How it works: Cosmic patching the AKS extension = a Helm upgrade. Setting controllerConfig.enabled=false re-renders the Deployment's pod-template env (CONTROLLER_ENABLED=false). Changing the pod template triggers a rolling restart; the new pods read the env at startup in [main()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and skip reconciler registration. Env is read once at boot, and the restart is what delivers the change.

Why that's okay:

Backward compatible. Defaults to true — existing installs behave exactly as today. It's additive and opt-in.
Safe blast radius when off. No reconcilers register → the controller takes no action on any namespace/PDB, but the manager still starts and serves health + metrics, so pods stay Ready — no crashloop, no probe failures, no eviction of the controller itself.
Reversible, no corruption. Flip back to true → another upgrade → rollout → reconcilers resume. The CRs use no finalizers, so disabling doesn't strand finalizers or wedge deletes.
Standard pattern. Config-driven, Helm-delivered, rolling-restart rollout is the normal way AKS extensions apply config — no custom hot-reload machinery to get wrong

cmd/main.go
```diff
@@ -209,58 +209,91 @@ func main() {
...
-		Scheme:            mgr.GetScheme(),
-		Filter:            nsfilter,
-		ZeroSurgeOverride: zeroSurgeOverride,
+	// Parse CONTROLLER_ENABLED environment variable (defaults to true). This is a
+	// global kill switch: when false, no reconcilers are registered, so the
+	// controller runs (serving health and metrics) but takes no action on any
+	// namespace or PDB — letting operators fully disable the feature in place via
+	// config, without uninstalling the extension.
+	controllerEnabled := true
+	if v := os.Getenv("CONTROLLER_ENABLED"); v != "" {
```
paulgmiller: discovered  this recently which is shorter. 

 k8senv "k8s.io/utils/env"

pdbCreate, err := k8senv.GetBool("PDB_CREATE", false)
+       if err != nil {
+               setupLog.Error(err, "Failed to parse PDB_CREATE env variable")
+               os.Exit(1)
        }

cmd/main.go
paulgmiller: in the pr I am working on Ipdb create goes away because i just pass the bool to the reconciler and always add it.  https://github.com/Azure/eviction-autoscaler/pull/175

NamanBhat: Done in `f59802c` — went with your slice-of-controllers + single-loop suggestion. Reconcilers are collected into a `[]reconcilerSetup` (a tiny interface for `SetupWithManager(mgr) error`), with the `pdbCreate` ones conditionally appended, then registered in one flat loop gated by the kill switch. Removes the deep nesting (net −19 lines).

cmd/main.go
NamanBhat: Nice, applied in `f59802c`: `CONTROLLER_ENABLED` now reads via `k8senv.GetBool("CONTROLLER_ENABLED", true)` (`k8s.io/utils/env`). Left `PDB_CREATE` alone since it goes away in your #175 — didn't want to churn code you're removing.

README.md
NamanBhat: Good point — these env vars are all read once at startup, so a ConfigMap/env change needs a pod restart to take effect. Heads up though: the kill-switch README docs (the `controllerConfig.enabled` bullet and the `CONTROLLER_ENABLED` env-var section) were removed from this PR in `d1da1fd` ("revert README documentation additions"), so the README no longer mentions the flag at all — there's nowhere in this PR to attach the startup-only note right now. Looks like the intent is to keep it undocumented here; if/when the docs come back, this caveat is worth including.

NamanBhat:
@paulgmiller — on the PR-description ask and the "won't the extension revert a direct edit?" question:

**Config-toggle vs uninstall:** flipping `controllerConfig.enabled=false` is an AKS-extension **config-settings patch** (`az k8s-extension update` / ARM `Update`) — the extension, CRDs, RBAC, webhooks, and all state (EAS CRs, PDBs, annotations) stay in place; only the manager pod rolls. Uninstall tears that down and loses in-place state, and re-enabling means a full reinstall + reconfigure. So it's one non-destructive ARM update vs a destructive uninstall/reinstall cycle.

**On reverting a direct edit — correct, and by design:** a `kubectl edit` of the Deployment would be reconciled back to the extension's desired state (drift correction). That's exactly why you flip it through **extension config settings**, not the Deployment — the change then originates from the extension's desired state (re-renders the pod-template env via Helm → rolling restart) and sticks. Same delivery path as every other setting (`ENABLED_BY_DEFAULT`, `ACTIONED_NAMESPACES`, `PDB_CREATE`).

I've added a "How it's toggled (and why not just uninstall)" section to the PR description capturing this.

skuchipudi295:
### Code review

skuchipudi295 (approved):

PR #163, opened 2026-08-05
permenantly conver max unavailable

paulgmiller:


cmd/main.go
```diff
@@ -195,6 +195,15 @@ func main() {
 	}
 	setupLog.Info("PDB creation configuration", "pdbCreate", pdbCreate)
 
+	maxUnavailableToMinAvailablePercentage, pdbConversionErr := controllers.ParseMaxUnavailableToMinAvailablePercentage(
```
paulgmiller, line 198: we could also make this match replicas with minAvailable but goes back to why are people setting maxunavailable in the first place. If they have 1000 replicas do they still want to go one by one?

cmd/main.go
```diff
@@ -246,9 +255,10 @@ func main() {
 	}
 
 	if err = (&controllers.PDBToEvictionAutoScalerReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
-		Filter: nsfilter,
+		Client:                                 mgr.GetClient(),
+		Scheme:                                 mgr.GetScheme(),
+		Filter:                                 nsfilter,
+		MaxUnavailableToMinAvailablePercentage: maxUnavailableToMinAvailablePercentage,
```
skuchipudi295, line 261: we would want to convert to minAvailble: int x not percent so we keep the floor same during surge or else it keeps moving

